PR #74, opened 2023-07-19
Set default seismogram output as ascii.

icui:
# Description

# Issue Number

# Checklist

[x] I ran the code through pre-commit to check style
[x] My code passes all the integration tests
[x] I have added sufficient unittests to test my changes
[x] I have added/updated documentation for the changes I am proposing
[x] I have updated CMakeLists to ensure my code builds
[x] My code builds across all platforms

### DIFF
--- a/include/domain/acoustic/acoustic_domain.tpp
+++ b/include/domain/acoustic/acoustic_domain.tpp
@@ -176,7 +176,7 @@ void initialize_sources(
 
     specfem::forcing_function::stf* source_time_function = compute_sources.h_stf_array(isource).T;
 
-    const auto kappax = Kokkos::subview(kappa, ispec, Kokkos::ALL(),
+    const auto sv_kappa = Kokkos::subview(kappa, ispec, Kokkos::ALL(),
                                  Kokkos::ALL());
 
     const auto sv_source_array =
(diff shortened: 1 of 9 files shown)


buildbot-princeton:

buildbot-princeton:

Rohit-Kakodkar:
ok to test

include/domain/acoustic/acoustic_domain.tpp
```diff
@@ -186,7 +187,7 @@ void initialize_sources(
             new (source)
                 source_type<qp_type,
                             specfem::enums::element::property::isotropic>(
-                    ispec, sv_source_array, source_time_function);
+                    ispec, kappa, sv_source_array, source_time_function);
```
Rohit-Kakodkar, line 190: I think kappa here should be of type_real and not a DeviceView. Since this is an elemental source class it would be better to pass to it the kappa of the element where this source lies. An analogy can be made to `source_array` for example where we pass only a subview of the `source_array` of the element where this source belongs. 

Advantage of doing this would be now if we ever wanted to test elemental source class by itself then we can just pass variables related to that element, resulting in self contained classes. In the current implementation  we'd have to pass kappa for the whole mesh, or at the very list create a scalar 3d view with value of the element - which seems counter intuitative

Rohit-Kakodkar (changes requested):
Minor organizational change on elemental sources class. Otherwise looks pretty good.

Issue #137, opened 2024-11-15
Pressure Seismogram


int-ptr-ptr:
**Is your feature request related to a problem? Please describe.**
A pressure seismogram is a feature in the original `SPECFEM` that would be beneficial to add in `SPECFEM++`.

**Describe the solution you'd like**
One should be able to add `-pressure` to `seismogram-type` in the configuration file, and produce seismogram outputs for pressure at stations.

**Additional context**
*Add any other context or screenshots about the feature request here.*

int-ptr-ptr:
For this to work, as well as making it easier for other seismogram types to be added, the current `compute_seismograms` routine may need to be reworked, since additional shared views may be needed.

Moreover, there is currently no good way to isolate code between seismogram types that need more than just a "use this field, instead"-type difference in implementation.

For our pressure case, we need to compute the divergence of the displacement field (or for the acoustic case, just take the acceleration field). I can see two ways of doing it:
- <span id="comment-implementation-pr138-initcommit"></span>As in commit 7e9d40bf949dc5708e1c8e418ddb8c61ba3f947d, we can add an auxiliary field within the receiver-seismogram `parfor` (requiring more scratch size), store the displacement field in there, then compute the divergence. This aux-field is used under the assumption that the receiver kernel's `get_field()` should be ignorant of the seismogram type.
- Alternatively, if we let `get_field()` access `seismogram_type_l`, then the divergence can be computed directly from the displacement field. Additionally, the acoustic kernel can just return the negative acceleration of $\\chi$ and skip the $\\chi$->displacement conversion.

PR #138, opened 2024-11-15
Pressure Seismogram

int-ptr-ptr:
# Description

This PR attempts to add a pressure seismogram option.

# Issue Number
#137

# Checklist

## Developers' List

Before certain checks in the [maintainers' list](#desc-maintainers-list-heading), the PR should be feature complete (ex. `CMakeLists` does not need an update at first, but a new feature in the PR creates a new `.tpp` file). These should be verified maintainer-side:

- [ ] <span id="desc-devlist-checkbox-filename"></span>The [writer](https://github.com/int-ptr-ptr/SPECFEMPP/blob/pressure_seismo/src/writer/seismogram.cpp) should have the right naming scheme. Commit 7e9d40bf949dc5708e1c8e418ddb8c61ba3f947d used `f"{network}{station}BX.semp"` as a placeholder.
- [ ] <span id="desc-devlist-checkbox-kernel"></span>The implementation in [kernel](https://github.com/int-ptr-ptr/SPECFEMPP/blob/pressure_seismo/include/domain/impl/receivers/kernel.tpp) should be satisfactory.

<h2 id="desc-maintainers-list-heading">Maintainers' List</h2>

### DIFF
--- a/include/domain/impl/receivers/kernel.tpp
+++ b/include/domain/impl/receivers/kernel.tpp
@@ -84,6 +84,10 @@ void specfem::domain::impl::kernels::receiver_kernel<
       NGLL, DimensionType, MediumTag, specfem::kokkos::DevScratchSpace,
       Kokkos::MemoryTraits<Kokkos::Unmanaged>, true, true, true, false, using_simd>;
 
+  using Aux2ComponentFieldType = specfem::element::field<
+      NGLL, DimensionType, specfem::element::medium_tag::elastic, specfem::kokkos::DevScratchSpace,
+      Kokkos::MemoryTraits<Kokkos::Unmanaged>, true, false, false, false, using_simd>;
+
   using ElementQuadratureType = specfem::element::quadrature<
       NGLL, DimensionType, specfem::kokkos::DevScratchSpace,
       Kokkos::MemoryTraits<Kokkos::Unmanaged>, true, false>;
@@ -95,7 +99,7 @@ void specfem::domain::impl::kernels::receiver_kernel<
     return;
 
   int scratch_size =
-      ElementFieldType::shmem_size() + ElementQuadratureType::shmem_size();
+      ElementFieldType::shmem_size() + ElementQuadratureType::shmem_size() + Aux2ComponentFieldType::shmem_size();
 
   Kokkos::parallel_for(
       "specfem::domain::domain::compute_seismogram",
@@ -118,6 +122,7 @@ void specfem::domain::impl::kernels::receiver_kernel<
         // ----------------------------------------------------------------
         ElementFieldType element_field(team_member);
         ElementQuadratureType element_quadrature(team_member);
+        Aux2ComponentFieldType aux_field(team_member);
 
         // Allocate shared views
         // ----------------------------------------------------------------
@@ -171,6 +176,9 @@ void specfem::domain::impl::kernels::receiver_kernel<
                     case specfem::enums::seismogram::type::acceleration:
                       return element_field.acceleration;
                       break;
+                    case specfem::enums::seismogram::type::pressure:
+                      return element_field.displacement;
+                      break;
                     default:
                       DEVICE_ASSERT(false, "seismogram not supported");
                       return {};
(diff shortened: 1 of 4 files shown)


buildbot-princeton:

buildbot-princeton:

buildbot-princeton:

src/writer/seismogram.cpp
```diff
@@ -43,6 +43,10 @@ void specfem::writer::seismogram::write() {
                        this->output_folder + "/" + network_name + station_name +
                            "BXZ" + ".sema" };
           break;
+        case specfem::enums::seismogram::type::pressure:
+          filename = { this->output_folder + "/" + network_name + station_name +
+                       "BX" + ".semp" }; // TODO get the right name
+          break;
```
int-ptr-ptr, line 49: Relevant to [this checklist](https://github.com/PrincetonUniversity/SPECFEMPP/pull/138#desc-devlist-checkbox-filename).

include/domain/impl/receivers/kernel.tpp
```diff
@@ -186,10 +194,110 @@ void specfem::domain::impl::kernels::receiver_kernel<
...
+                if constexpr (MediumTag == specfem::element::medium_tag::acoustic){
+                  return point_properties.kappa;
+                }else if constexpr (MediumTag == specfem::element::medium_tag::elastic){
+                  //we may also need an isotropic if statement?
+                  return point_properties.lambdaplus2mu;
+                }else{
+                  static_assert(false, "include/domain/impl/receivers/kernel.tpp: unknown medium to retrieve kappa!");
+                  return 0;
+                }
+              }();
```
int-ptr-ptr, line 256: <span id="comment-bulk-modulus-params-code"></span>These may or may not be the right properties for the bulk modulus. Please verify.

int-ptr-ptr: ([link to return to checklist](#desc-devlist-checkbox-bulk_modulus))

Rohit-Kakodkar:
I think your formulation is different then what has been previously used in SPECFEM2D.

For elastic domains - https://github.com/SPECFEM/specfem2d/blob/98741db1a0c8082ca57364f5b17ea95df6cbf1c2/src/specfem2D/compute_pressure.f90#L398

For acoustic domains - https://github.com/SPECFEM/specfem2d/blob/98741db1a0c8082ca57364f5b17ea95df6cbf1c2/src/specfem2D/compute_pressure.f90#L569

I think we should use the same formulations.

include/domain/impl/receivers/kernel.tpp
```diff
@@ -84,6 +84,10 @@ void specfem::domain::impl::kernels::receiver_kernel<
       NGLL, DimensionType, MediumTag, specfem::kokkos::DevScratchSpace,
       Kokkos::MemoryTraits<Kokkos::Unmanaged>, true, true, true, false, using_simd>;
 
+  using Aux2ComponentFieldType = specfem::element::field<
```
Rohit-Kakodkar, line 87: I think what you might want to use is `using Aux2ComponentFieldType = typename ElementFieldType::ViewType`. This should directly give you access to the underlying Kokkos::View used to store the data

include/domain/impl/receivers/kernel.tpp
```diff
@@ -186,10 +194,110 @@ void specfem::domain::impl::kernels::receiver_kernel<
...
+              if(seismogram_type_l == specfem::enums::seismogram::type::pressure){
+                //for pressure, we need to compute -kappa * div(s)
+                //so we store displacement into auxfield so we can overwrite
+                //sv_receiver_field
+
+                #ifndef KOKKOS_ENABLE_CUDA
+                #pragma unroll
+                #endif
+                for (int l = 0; l < specfem::dimension::dimension<DimensionType>::dim; l++) {
+                  aux_field.displacement(iz,ix,l) = sv_receiver_field(l);
```
Rohit-Kakodkar, line 206: With direct access to Kokkos::View as stated in the comment above. This could be `aux_field(iz, ix, l)`

include/domain/impl/receivers/kernel.tpp
```diff
@@ -186,10 +194,110 @@ void specfem::domain::impl::kernels::receiver_kernel<
                                  point_properties,
```
Rohit-Kakodkar, line 194: I think it would be ideal if we can pass seismogram type to this function. I realize for pressure we need the seismo type within `get_field`. With your current formulation this seems not possible, since we need the divergence. Can you look at the SPECFEM2D formulation https://github.com/SPECFEM/specfem2d/blob/master/src/specfem2D/compute_pressure.f90 .

With that formulation we should be able to compute pressure using only gradients

src/writer/seismogram.cpp
Rohit-Kakodkar: The naming scheme is based off SPECFEM2D seismogram naming scheme. https://github.com/SPECFEM/specfem2d/blob/98741db1a0c8082ca57364f5b17ea95df6cbf1c2/src/specfem2D/write_seismograms.F90#L302

int-ptr-ptr:
I've added seismogram-typing to `get_field` parameters, and moved `active_field`-setting to `get_field`.

As for unit testing, I was thinking about just having a set of (SPECFEM++ parameter-file, SPECFEM2D seismogram-output-files) pairs to compare against. The current seismogram tests don't have a ground truth for pressure.

Rohit-Kakodkar:
> I've added seismogram-typing to `get_field` parameters, and moved `active_field`-setting to `get_field`.
> 
> As for unit testing, I was thinking about just having a set of (SPECFEM++ parameter-file, SPECFEM2D seismogram-output-files) pairs to compare against. The current seismogram tests don't have a ground truth for pressure.

Yes I think thats the best way of doing it right now. Where we compare seismograms from the 2 versions of the software. We already do that within displacement tests  (https://github.com/PrincetonUniversity/SPECFEMPP/blob/78365e29dca1ad784f336bdefdbd5b370585fb69/tests/unit-tests/displacement_tests/Newmark/newmark_tests.cpp#L283). In that case I compare displacement seismograms, but you could easily do the same by adding pressure seismograms in `traces` folder.

include/domain/impl/receivers/elastic/elastic2d_isotropic.tpp
```diff
@@ -43,23 +43,74 @@ KOKKOS_INLINE_FUNCTION void specfem::domain::impl::receivers::receiver<
...
+#endif
+    for (int l = 0; l < NGLL; l++) {
+      dsx_dxi += hprime(ix, l) * element_field.displacement(iz, l, 0);
+      dsx_dgamma += hprime(iz, l) * element_field.displacement(l, ix, 0);
+      dsz_dxi += hprime(ix, l) * element_field.displacement(iz, l, 1);
+      dsz_dgamma += hprime(iz, l) * element_field.displacement(l, ix, 1);
+    }
+
+    // -kappa * (dsxdx + dszdz)
+    receiver_field(0) = -properties.lambdaplus2mu * (
```
Rohit-Kakodkar, line 105: Seems like your computation for pressure field is different from the original formulation. In the original formulation pressure = trace of stress tensor (https://github.com/SPECFEM/specfem2d/blob/98741db1a0c8082ca57364f5b17ea95df6cbf1c2/src/specfem2D/compute_pressure.f90#L398)

include/domain/impl/receivers/elastic/elastic2d_isotropic.tpp
```diff
@@ -43,23 +43,74 @@ KOKKOS_INLINE_FUNCTION void specfem::domain::impl::receivers::receiver<
...
+      (dsx_dxi * partial_derivatives.xix +
```
Rohit-Kakodkar, line 106: Also seems like I havent given access to kappa within the point properties type. lambdaplus2mu != kappa (https://github.com/PrincetonUniversity/SPECFEMPP/blob/78365e29dca1ad784f336bdefdbd5b370585fb69/include/material/elastic_properties.hpp#L45).

include/domain/impl/receivers/elastic/elastic2d_isotropic.tpp
```diff
@@ -43,23 +43,110 @@ KOKKOS_INLINE_FUNCTION void specfem::domain::impl::receivers::receiver<
...
+    dsz_dx = dsz_dxi    * partial_derivatives.xix +
+             dsz_dgamma * partial_derivatives.gammax;
+    dsz_dz = dsz_dxi    * partial_derivatives.xiz +
+             dsz_dgamma * partial_derivatives.gammaz;
+
+    //https://specfem2d-kokkos.readthedocs.io/en/devel/api/datatypes/field_derivatives/point.html#_CPPv4N7specfem5point17field_derivatives2duE
+    // tells us that fhe first index is the derivative index
+#define du(i,j) (i==0? (j==0? dsx_dx:dsz_dx):(j==0? dsx_dz:dsz_dz))
+
+    // WET code... should this be refactored with
```
Rohit-Kakodkar, line 117: I agree. At later point, we should just call the function from impl/elements. I think I need to do some refactoring before that. For now this implementation looks good.

include/domain/impl/receivers/elastic/elastic2d_isotropic.tpp
```diff
@@ -43,23 +43,110 @@ KOKKOS_INLINE_FUNCTION void specfem::domain::impl::receivers::receiver<
...
+    dsx_dz = dsx_dxi    * partial_derivatives.xiz +
+             dsx_dgamma * partial_derivatives.gammaz;
+    dsz_dx = dsz_dxi    * partial_derivatives.xix +
+             dsz_dgamma * partial_derivatives.gammax;
+    dsz_dz = dsz_dxi    * partial_derivatives.xiz +
+             dsz_dgamma * partial_derivatives.gammaz;
+
+    //https://specfem2d-kokkos.readthedocs.io/en/devel/api/datatypes/field_derivatives/point.html#_CPPv4N7specfem5point17field_derivatives2duE
+    // tells us that fhe first index is the derivative index
+#define du(i,j) (i==0? (j==0? dsx_dx:dsz_dx):(j==0? dsx_dz:dsz_dz))
```
Rohit-Kakodkar, line 115: This LGTM, we should eventually define this as field_Derivatives. I think some refactor is needed before then. For now this is fine.

include/domain/impl/receivers/elastic/elastic2d_isotropic.tpp
```diff
@@ -43,23 +43,110 @@ KOKKOS_INLINE_FUNCTION void specfem::domain::impl::receivers::receiver<
...
+                                          // CPU-version
+
+      // sigma_xz
+      sigma_xz = properties.mu * du(1, 0); // sigma_zy
+    }
+
+#undef du
+    // https://github.com/SPECFEM/specfem2d/blob/98741db1a0c8082ca57364f5b17ea95df6cbf1c2/src/specfem2D/compute_pressure.f90#L398
+    // pressure = - trace(sigma) / 3
+    receiver_field(0) = - (sigma_xx + sigma_zz)/3.0;
```
Rohit-Kakodkar, line 146: You missed `sigma_yy` which in 2D is `lambdaplus2mu * (du(0,0) + du(1,1))`

int-ptr-ptr: Thank you for catching that! I was planning on looking for a lost factor of 0.75 after the meeting. This might be it.

int-ptr-ptr:
The `database.bin` files from the most recent version of the Fortran code error out when being read in to SPECFEM++, so I am still using the old `database.bin` files.

It seems like serial test 3 has the wrong displacement, but serial tests 1 and 2 are good for all 4 seismogram types. It does not seem like the provenance parfile generates the same simulation.

Serial tests 5-7 have acceleration relative errors ~ 5e-2, and there are some stations that give all zeros for pressure Fortran-side, so rel-errors are failing out as `NaN`.

include/domain/impl/receivers/kernel.tpp
Rohit-Kakodkar: Can we remove `Aux2ComponentFieldType` if its not getting used?

Rohit-Kakodkar:
> The `database.bin` files from the most recent version of the Fortran code error out when being read in to SPECFEM++, so I am still using the old `database.bin` files.
> 
> It seems like serial test 3 has the wrong displacement, but serial tests 1 and 2 are good for all 4 seismogram types. It does not seem like the provenance parfile generates the same simulation.
> 
> Serial tests 5-7 have acceleration relative errors ~ 5e-2, and there are some stations that give all zeros for pressure Fortran-side, so rel-errors are failing out as `NaN`.

It might not be the best to test all seismograms in all simulations then. Especially if the seismogram values are very small. I think that might be the issue with acceleration. I am not sure what issue you are seeing with serial test 3, unless the reference Fortran seismograms have changed. The implementation in this PR hasnt modified how displacement seismograms should be computed. 

If you selectively wanted to add pressure seimogram tests, I would recommend adding it to serial test 1 and 2 since it'll test both elastic and acoustic implementation individually.

Rohit-Kakodkar:
Also I am used git commit ID `5bacf50dcd9c295cd236c8fb602006f9ac97b107` for the fortran code. i.e. thats where my devel is. I think seismoelectric domains have been added since then, which will cause errors when reading the database files

int-ptr-ptr:
To add/remove seismgrams to the tests, it suffices to modify the parameter `.yaml` file. Tests 1 and 2 currently check all 4 types, but test 3 onward only look at velocity. I've also reverted the traces on test 3, since mine were different. I think we're ready for integration/build tests.

Is the specfem parameter documentation the only part of the docs that has seismograms? I added a table [here](https://specfem2d-kokkos--138.org.readthedocs.build/en/138/parameter_documentation/receivers.html). Is that sufficient, or is there somewhere else that needs changing?

lsawade:
This is likely unrelated to you. The test keep on failing because they are run in parallel and access the same repo dir. We're working on it!

PR #186, opened 2024-11-26
Hello world for pybind

icui:
# Description

Adds a python module when performing ```pip install```. The python package is located in python/specfempp, which follows the convention for pybind. In the future we can change this folder into a submodule.

# Issue Number

#182 

# Checklist

- [x] My code builds across all platforms


Rohit-Kakodkar (commented):
Why do we need to create a seperate driver file `specfem2d_py.cpp`? Can we not add the bindings to the same file?

CMakeLists.txt
```diff
@@ -15,13 +15,26 @@ set(NUM_CHUNKS 1)
...
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
+            set(CMAKE_CXX_FLAGS "-fp-model=precise -fPIC")
+    elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
+            message("-- Detected Intel classic compiler which will be deprecated soon.")
+            message("-- It is recommended you use IntelLLVM compiler.")
+            set(CMAKE_CXX_FLAGS "-diag-disable=10441 -fp-model=precise -fPIC")
+    else()
+            set(CMAKE_CXX_FLAGS "-fPIC")
+    endif()
+else (SPECFEMPP_ENABLE_PYTHON)
```
Rohit-Kakodkar, line 29: Is this syntax correct?

icui:
> Why do we need to create a seperate driver file `specfem2d_py.cpp`? Can we not add the bindings to the same file?
Fixed now.

CMakeLists.txt
icui: I think so. I have tested with both pip and cmake.

CMakeLists.txt
```diff
@@ -534,7 +547,35 @@ else (DOXYGEN_FOUND)
...
-if (SPECFEMPP_ENABLE_PYTHON)
+if (SPECFEMPP_USE_SKBUILD)
     install(TARGETS specfem2d DESTINATION ${SKBUILD_SCRIPTS_DIR})
     install(FILES ${CMAKE_BINARY_DIR}/xmeshfem2D DESTINATION ${SKBUILD_SCRIPTS_DIR} PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE)
+endif (SPECFEMPP_USE_SKBUILD)
+
+if (SPECFEMPP_ENABLE_PYTHON)
+    find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
+    find_package(pybind11 CONFIG REQUIRED)
+    python_add_library(_core MODULE src/specfem2d.cpp WITH_SOABI)
```
Rohit-Kakodkar, line 558: Is this just taking the bindings from `src/specfem2d.cpp` and then compiling it under `_core`

icui: Yes

Rohit-Kakodkar (commented):
> > Why do we need to create a seperate driver file `specfem2d_py.cpp`? Can we not add the bindings to the same file?

Fixed now. 

Eventually we'd want to move bindings to a `driver` module that runs specfem. And call the driver in main

Rohit-Kakodkar (commented):
Otherwise this looks good

CMakeLists.txt
lsawade: I actually like this syntax. It clears up what conditional you are in, and nothing would be put after else anyways.

I haven't seen it like that either though.

lsawade (approved):
Looks good to me

PR #196, opened 2024-11-27
moved read_mesh/sources/receivers to the IO/interface.hpp

lsawade:
# Description

# Issue Number

# Checklist


CMakeLists.txt
```diff
@@ -129,15 +129,13 @@ add_library(
         src/IO/fortranio/fortran_io.cpp
         src/IO/HDF5/native_type.cpp
         # src/IO/ASCII/native_type.cpp
-        src/IO/mesh/read_mesh.cpp
+        src/IO/interface.cpp
```
Rohit-Kakodkar, line 132: Thoughts on keeping the `cpp` files separate. i.e. `src/IO/read_mesh.cpp` `src/IO/read_sources.cpp` and `src/IO/read_receivers.cpp` but having a single declaration file `src/IO/interface.hpp`.

lsawade: Would be cleaner. I could rename them `src/IO/receivers.cpp`, `src/IO/receivers.cpp`, `src/IO/mesh.cpp`, it would be more elegant than having a giant file with the definitions.

lsawade: Will revert the source files, good point. Nothing will change except the CMakeLists.txt.

Rohit-Kakodkar (commented):
> Will revert the source files, good point. Nothing will change except the CMakeLists.txt. 

Sounds good. Otherwise the PR LGTM.

PR #217, opened 2024-12-03
Starting a doc to track the implementation of new physics.

lsawade:
# Description

Added `docs/implement_new_physics.rst` to track which files need to be updated and changed to add new things.


Rohit-Kakodkar:
I think we should create an issue for anisotropic domains and keep this within that issue branch. We can expand this doc into a tutorial before merging into devel

PR #220, opened 2024-12-03
Discontinuous Galerkin (dG)

int-ptr-ptr:
# Description

_Please describe the changes/features in this pull request._ 

This PR is the (WIP) implementation for dG / mortar method.

As of the current stage of development, we are utilizing an event system pattern, where events are queued as (function, float) pairs, which I am calling _events_, where the functions are called in order of increasing floating point value, effectively serving as a program with a continuous, rather than integer, program counter. The current implementation is rather bare-bones as things like interrupts and stack traces are unneeded for my purposes at the moment.

# Issue Number

_If there is an issue created for these changes, link it here_

# TODOs
- [ ] Complete and test the different medium interfaces
  - [x] fluid-fluid
  - [ ] fluid-solid
  - [ ] solid-solid
- [ ] Redo the edge container
  - [ ] Somehow template edges on medium, etc. so that `edge.data` is not the mess that it is now
  - [ ] add `..._on_device` methods
  - [ ] divorce the Kokkos views from the edge structs (edge structs could be populated by views, as in `PointViewType`-like structs)
  - [ ] parallelize the `foreach` methods -- we could keep `std::function` args as an `..._on_host` overload, but some way of having an inline-able Kokkos function is probably necessary.
- [ ] Improve the event system (maybe break into its own PR to work with pybind, etc?)

# Checklist

_Please make sure to check developer documentation on specfem docs._


buildbot-princeton:

buildbot-princeton:

buildbot-princeton:

PR #271, opened 2024-12-10
Added the anisotropic medium header

### LINKED ISSUE #270: Create `include/medium/elastic_anisotropic2d/elastic_anisotropic2d.hpp`

lsawade:
# Description

adds `specfem/medium/elastic_anisotropic2d/elastic_anisotropic2d.hpp`

# Issue Number

closes #270 

# Checklist


lsawade:
I specifically did not include it in medium/compute_{stress, wavefield}.hpp since that needs to be updated.

icui:
specfem2d has a safeguard that prohibits calculating pressure when c12 or c23 is zero. Should we add that?
```cpp
        if (c12 < 1.e-7 || c23 < 1.e-7){
          // cannot compute pressure for an anisotropic material if c12 or c23 are zero
          sigma_xx = 0.0f; sigma_zz = 0.0f; sigma_yy = 0.0f;
        }
```

Rohit-Kakodkar:
> specfem2d has a safeguard that prohibits calculating pressure when c12 or c23 is zero. Should we add that?
> 
> ```c++
>         if (c12 < 1.e-7 || c23 < 1.e-7){
>           // cannot compute pressure for an anisotropic material if c12 or c23 are zero
>           sigma_xx = 0.0f; sigma_zz = 0.0f; sigma_yy = 0.0f;
>         }
> ```

I support this. I'd add it before line 97 so as to avoid gradient calculation. And return `wavefield(:, :, :, 0) = 0` <- not a valid syntax btw. Youll need a `Kokkos::parallel_for` loop

Rohit-Kakodkar (approved):
Apart from @icui comment ii LGTM

icui (approved):

lsawade:
@Rohit-Kakodkar 

Ok I rechecked the code for the wavefield check, throwing an error is easy enough -- so can do. But I'm also not clear on the nested `Kokkos::parallel_for` after rereading the code and thinking about your suggestion. 

inside the 
```
if compute_pressure==True
```
there was originally
```
compute_gradient(..., medium specific callback)  {
    parallel_for (
            compute spatial derivative
            run medium specific callback
    )
```
This runs of course over elements and all points. Your suggestion was to put
```
parallel_for (
        if (c12 < 1.e-7 || c23 < 1.e-7) {
          wavefield(...) = 0.0
        }
);
return;
```
But this computes this for all elements, some of which may not have c12 or c23 for which this is valid.
and so a return after would not make sense as the gradient computation would still be required for the locations, at which the condition is false. I can't put this in the callback either since we would want to avoid gradient computation, but that happens before the callback is run anyways.

Am I missing something in the kokkos syntax here/ did I misunderstand your suggestion?

Will update the program to throw and error for now.

Rohit-Kakodkar:
> @Rohit-Kakodkar
> 
> Ok I rechecked the code for the wavefield check, throwing an error is easy enough -- so can do. But I'm also not clear on the nested `Kokkos::parallel_for` after rereading the code and thinking about your suggestion.
> 
> inside the
> 
> ```
> if compute_pressure==True
> ```
> 
> there was originally
> 
> ```
> compute_gradient(..., medium specific callback)  {
>     parallel_for (
>             compute spatial derivative
>             run medium specific callback
>     )
> ```
> 
> This runs of course over elements and all points. Your suggestion was to put
> 
> ```
> parallel_for (
>         if (c12 < 1.e-7 || c23 < 1.e-7) {
>           wavefield(...) = 0.0
>         }
> );
> return;
> ```
> 
> But this computes this for all elements, some of which may not have c12 or c23 for which this is valid. and so a return after would not make sense as the gradient computation would still be required for the locations, at which the condition is false. I can't put this in the callback either since we would want to avoid gradient computation, but that happens before the callback is run anyways.
> 
> Am I missing something in the kokkos syntax here/ did I misunderstand your suggestion?
> 
> Will update the program to throw and error for now.

Okay I did not think of that case. Yes in that case you'd have to do the computation of doing the gradient. No way around it under the current framework. But as @icui said throwing an error makes more sense.

include/medium/elastic_anisotropic2d/elastic_anisotropic2d.hpp
```diff
@@ -0,0 +1,163 @@
...
+        const auto iterator_index = iterator(i);
+        const auto &index = iterator_index.index;
+
+        // Load the properties
+        PointPropertyType point_property;
+        specfem::compute::load_on_device(index, properties, point_property);
+
+        // cannot compute pressure for an anisotropic material if c12 or c23 are zero
+        if (point_property.c12 < 1.e-7 || point_property.c23 < 1.e-7) {
+          Kokkos::abort("C_12 or C_23 are zero, cannot compute pressure. Check your material properties. Or, deactivate the pressure computation.");
```
Rohit-Kakodkar, line 110: I think you forgot to run the pre-commit hooks here

PR #348, opened 2024-12-26
Property reader / writer

### LINKED ISSUE #343: property reader

icui:
# Description

Read / write GLL properties. Usage:
```yaml
  ## databases
  databases:
    mesh-database: "./OUTPUT_FILES/database.bin"
    source-file: "./source.yaml"
    reader:
      properties:
        directory: GLL
        format: ASCII
    writer:
      properties:
        directory: GLL
        format: ASCII
```

# Issue Number

Closes #343 and #344


lsawade (approved):
I think a test should be implemented to check whether reading/writing is done properly, other than that it L Amazing TM!

src/specfem2d.cpp
```diff
@@ -163,9 +163,31 @@ void execute(const std::string &parameter_file, const std::string &default_file,
   specfem::compute::assembly assembly(
       mesh, quadrature, sources, receivers, setup.get_seismogram_types(),
       setup.get_t0(), dt, nsteps, max_seismogram_time_step,
-      setup.get_simulation_type());
+      setup.get_simulation_type(), true);
```
Rohit-Kakodkar, line 166: The assign property boolean should come from setup class. If we are planning to set it to true always then we dont need a boolean

Rohit-Kakodkar (changes requested):

src/specfem2d.cpp
icui: Thanks, I forgot to change this. Now updated.

Rohit-Kakodkar (approved):
I think we saw these changes as previous PR into `issue-343`

PR #356, opened 2024-12-31
Issue 355 -- Fixing docker image creation

### LINKED ISSUE #355: Docker build failing on devel branch
**Describe the bug**

After adding meshfem3d, the docker build is failing. This is likely due to an old version of CMake being used. `INTERFACE` libraries are supported from `cmake=3.19` and the docker publishing is using `cmake=3.18.4`

**To Reproduce**

Use `cmake=3.18` locally reproduces the error.

lsawade:
# Description

Manually installs cmake, since the Debian repo is not installing version>=3.19.

# Issue Number

closes #355

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -27,7 +27,7 @@ jobs:
         with:
           images: ghcr.io/PrincetonUniversity/SPECFEMPP
           tags: |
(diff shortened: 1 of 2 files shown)


lsawade:
happy to adapt to an apt-get/apt solution, but that does not seem to work...

Rohit-Kakodkar (approved):
Nice! Looks a very annoying fix. Thanks for fixing this.

icui (approved):

PR #358, opened 2025-01-02
Issue 357 - Fix Boost installation to remove non-official dependency

### LINKED ISSUE #357: Boost has new hash URL
**Describe the bug**

The current installation of Boost requires a hash, which now has become outdated. Furthermore, the current boost build is not from the 
official boost repo.

**To Reproduce**

Install without existing boost installation.

**Expected behavior**

Boost is downloaded and installed

**Machine information (please complete the following information):**

Machine independent

lsawade:
# Description

The hash for the boost download changed, which is a dependency that is outside of this repo that could change anytime again. This pull request introduces a fast way to build  boost from the official repo's release zips, only building the required libraries.

It should fix previous build issues.

# Issue Number

Closes #357 

# Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,26 +105,26 @@ endif()
 
 
 # Try finding boost and if not found install.
-find_package(Boost 1.83.0 COMPONENTS program_options filesystem system)
+find_package(Boost 1.78.0 COMPONENTS program_options filesystem system)
 
 if (NOT ${Boost_FOUND})
     # Add boost lib sources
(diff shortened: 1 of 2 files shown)


lsawade:
Boost seems to be installing correctly (Looked at the running logs). I did not use a Generator expression since it seemed unnecessarily ugly when I already have a giant conditional thing to install Boost. 
So, `BOOST_LIBS` now contains all the boost dependencies that are required from Boost for the local della installation or the ones required by the manual installation of the most recent version.
It is not clean, but I think that is related to the della installation rather than the general case.

icui (approved):

PR #359, opened 2025-01-02
Issue 319 -- mass matrix 

### LINKED ISSUE #319: Move `specfem::domain::impl::element::impl_mass_matrix` to `specfem::medium::mass_matrix`

lsawade:
# Description

# Issue Number

closes #319 

# Checklist


src/medium/compute_mass_matrix.cpp
```diff
@@ -0,0 +1,47 @@
+#include "medium/compute_mass_matrix.hpp"
+
+// This file only contains explicit template instantiations
+
+// dim2, elastic, isotropic, false
+template specfem::point::field<specfem::dimension::type::dim2,
```
Rohit-Kakodkar, line 6: Do you want to use macros for template instantiation here?

include/medium/compute_source.hpp
```diff
@@ -1,8 +1,8 @@
 #pragma once
 
-#include "acoustic_isotropic2d/source.hpp"
-#include "elastic_anisotropic2d/source.hpp"
-#include "elastic_isotropic2d/source.hpp"
+#include "2d/acoustic/isotropic/source.hpp"
```
Rohit-Kakodkar, line 3: Can we change this to "dim2" rather than "2d"? File names generally should not start with numbers

Rohit-Kakodkar (changes requested):

src/medium/compute_mass_matrix.cpp
```diff
@@ -0,0 +1,30 @@
+#include "medium/compute_mass_matrix.hpp"
+#include "enumerations/dimension.hpp"
+#include "enumerations/material_definitions.hpp"
+#include "enumerations/medium.hpp"
+
+// This file only contains explicit template instantiations
+
+#define INSTANTIATION_MACRO(DIMENSION_TAG, MEDIUM_TAG, PROPERTY_TAG)           \
+  template specfem::point::field<DIMENSION_TAG, MEDIUM_TAG, false, false,      \
```
Rohit-Kakodkar, line 9: The instantiation macro now needs GET_TAG macro get the tag. i.e. `template specfem::point::field<GET_TAG(DIMENSION_TAG), GET_TAG(MEDIUM_TAG), false, false, ...`

Rohit-Kakodkar: See https://github.com/PrincetonUniversity/SPECFEMPP/blob/901911af65325f9be540546e23ff4fefa9d9c62f/include/enumerations/material_definitions.hpp#L339

lsawade: Was missing the skill to properly copy from `kernel/frechet_kernel.cpp` ...

lsawade:
NOT COMPILING ON GPU

lsawade:
Ok this is weird. Only for GPU the the code does not compile throwing following error:
```
/home/lsawade/SPECFEMPP/include/enumerations/material_definitions.hpp:115:1: error: macro "BOOST_PP_TUPLE_ELEM" requires 3 arguments, but only 2 given
```
This is only happening for the GPU compilation, so I don't think the issue is `boost` in this case. 

But if I do not load `boost` on the cluster, and compile it as part of the build process, a different error unrelated to boost is thrown.
```
/home/lsawade/SPECFEMPP/include/kernels/frechet_kernels.tpp(76): error: The enclosing parent function ("compute_material_derivatives") for an extended __host__ __device__ lambda cannot have private or protected access within its class
        [ =, *this ] __attribute__((host)) __attribute__((device))(const typename ChunkPolicy::member_type &team) {
                                                                  ^
```

lsawade:
Further investigation shows that the commit [f52ad23](https://github.com/PrincetonUniversity/SPECFEMPP/commit/f52ad231bc5b7a1c3e89f1f4799d1e6ff1055193) is not compiling either using cluster installed `boost`. Which was the original move of the frechet derivatives to medium. Maybe you could take a look @Rohit-Kakodkar ?

Rohit-Kakodkar:
> Further investigation shows that the commit [f52ad23](https://github.com/PrincetonUniversity/SPECFEMPP/commit/f52ad231bc5b7a1c3e89f1f4799d1e6ff1055193) is not compiling either on GPU using cluster installed `boost`. Which was the original move of the frechet derivatives to medium. Maybe you could take a look @Rohit-Kakodkar ?

Yes I will have a look at this.

Rohit-Kakodkar (approved):
You can merge this into the 228 branch. I'll add a fix in another issue

lsawade:
Sweet!

PR #433, opened 2025-01-24
Separate Python code from specfem2d.cpp

### LINKED ISSUE #432: Separate Python code from specfem2d.cpp

icui:
# Issue Number

Closes #432


include/execute.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <vector>
+// Specfem2d driver
+
+void execute(
+    const YAML::Node &parameter_dict, const YAML::Node &default_dict,
+    std::vector<std::shared_ptr<specfem::periodic_tasks::periodic_task> > tasks,
+    specfem::MPI::MPI *mpi);
```
lsawade, line 28: should this maybe be specfem.hpp? since it'll contain all the headers required for running a simulation?

src/python/core.cpp
```diff
@@ -0,0 +1,105 @@
+#include "execute.hpp"
```
lsawade, line 1: Then this would call the specfem header

src/specfem2d.cpp
```diff
@@ -1,56 +1,4 @@
...
-          << " secs\n"
-          << "Simulation end time : " << ctime(&c_now)
-          << "------------------------------------------------\n";
-
-  return message.str();
-}
+#include "execute.hpp"
 
 boost::program_options::options_description define_args() {
   namespace po = boost::program_options;
```
lsawade, line 4: This file could keep its name still

CMakeLists.txt
```diff
@@ -627,6 +625,18 @@ target_link_libraries(
...
+add_executable(
+        specfem2d
+        src/specfem2d.cpp
+)
+
+
+
+target_link_libraries(
+        specfem2d
+        execute
```
lsawade, line 637: then the executable specfem2d

lsawade (approved):
Minor comments on the name, but am not married to suggestions. Otherwise LGTM

CMakeLists.txt
icui: specfem2d.cpp contains main(), which is not needed by Python. execute.cpp contains only the execute() function.

lsawade: I think there is text missing in this comment (did it from vscode). What I meant is: then the specfem2d executable and python/core are dependent on specfem.h/cpp

PR #435, opened 2025-01-24
Move all print statements to execute()

### LINKED ISSUE #434: Move all print statements to execute()
Right now some print statements are called automatically by instantiate() and some are called by execute(). Moving them all to execute() makes the behavior more consistent.

 - `quadrature.instantiate`
 - `read_mesh`

icui:
# Issue Number

Closes #434

### DIFF
--- a/include/quadrature/quadratures.hpp
+++ b/include/quadrature/quadratures.hpp
@@ -9,7 +9,18 @@ struct quadratures {
   specfem::quadrature::gll::gll gll;
 
   quadratures(const specfem::quadrature::gll::gll &gll) : gll(gll) {}
+
+  void print(std::ostream &out) const {
+    out << " Quadrature in X-dimension \n";
+    gll.print(out);
+
+    out << " Quadrature in Z-dimension \n";
+    gll.print(out);
+  }
 };
+
+std::ostream &operator<<(std::ostream &out,
+                         const specfem::quadrature::quadratures &quad);
 } // namespace quadrature
 } // namespace specfem
 
--- a/src/IO/mesh.cpp
+++ b/src/IO/mesh.cpp
@@ -185,7 +185,27 @@ specfem::IO::read_mesh(const std::string filename,
 
   stream.close();
 
-  // Print material properties
+  const auto &l_elastic_isotropic =
+      mesh.materials.elastic_isotropic.material_properties;
+  const auto &l_acoustic_isotropic =
+      mesh.materials.acoustic_isotropic.material_properties;
+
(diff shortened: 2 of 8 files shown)


src/IO/mesh.cpp
```diff
@@ -185,20 +185,40 @@ specfem::IO::read_mesh(const std::string filename,
 
   stream.close();
 
-  // Print material properties
+  const auto &l_elastic_isotropic =
```
lsawade, line 188: Instead of a separate print mesh function, the osstream operator could be overloaded like so:
```cpp
class A {
    int val;
public:
    A(int v) : val(v) {}
    friend ostream& operator<<(ostream& os, const A& obj) {
        os << obj.val;
        return os;
    }
};
``` 

then you can just 
```cpp
std::cout << mesh << std::endl;
```

include/quadrature/quadratures.hpp
```diff
@@ -9,7 +9,18 @@ struct quadratures {
...
 
   quadratures(const specfem::quadrature::gll::gll &gll) : gll(gll) {}
+
+  void print(std::ostream &out) const {
+    out << " Quadrature in X-dimension \n";
+    gll.print(out);
+
+    out << " Quadrature in Z-dimension \n";
+    gll.print(out);
+  }
```
lsawade, line 19: similarly here

src/execute.cpp
```diff
@@ -41,8 +41,15 @@ void execute(
   // --------------------------------------------------------------
   //                   Read mesh and materials
   // --------------------------------------------------------------
-  const auto quadrature = setup.instantiate_quadrature();
+  const auto quadratures = setup.instantiate_quadrature();
+  if (mpi->main_proc()) {
```
lsawade, line 45: I think the options are two, either you check here whether it's the main_proc() and then use the ostream << overload, or you parse mpi check locally if you are on main_proc and return if you aren't.

What do you think? Am i forgetting something?

Option 3:
the struct should all have the osstream overload. mpi->cout should really check whether you are on main proc.

lsawade (approved):
I like this otherwise

src/IO/mesh.cpp
icui: cout << mesh is already defined and prints different stuff

include/quadrature/quadratures.hpp
icui: std::cout << gll has already been defined, but at this stage not declared. I could use some help on the import order.

PR #448, opened 2025-02-04
Gpu compilation fix

lsawade:
# Description

final fixes for intel compiler


lsawade:
I'll merge this once it's passed the basic tests so that the anisotropic tests can run

PR #503, opened 2025-02-21
Issue 502 - Implements new read_array templates dim3 absorbing boundaries write/read

### LINKED ISSUE #502: Read absorbing boundaries and free surface

lsawade:
## Description

This PR implements 

- Further `Kokkos::View` reads of 2D and 3D views for `read_array` and `read_index_array`.
- Writing function in the Fortran code to write boundary indeces to write/read faces in the first dimension
- absorbing boundary struct
- templated stacey boundary mass matrices as part of the absorbing boundary struct
- free surface boundary struct
- reading of the free surface
- refactor of the reading catching errors to print information on the buffers
- function to read check values and printing the associated error (if there is one)

## Issue Number

Closes #502 

## Checklist


fortran/meshfem3d/generate_databases/save_arrays_solver.F90
```diff
@@ -221,10 +291,17 @@ subroutine save_arrays_solver_mesh()
   ! write(IOUT) kappastore
   ! write(IOUT) mustore
 
-  write(IOUT)
+
+  ! Converting the logical arrays to integer arrays and writing them as ints
+  ! call save_logical_nspec_array(ispec_is_acoustic)
```
Rohit-Kakodkar, line 296: Should this be commented?

include/IO/mesh/impl/fortran/dim3/interface.hpp
```diff
@@ -60,12 +60,20 @@ void read_partial_derivatives(
...
 
+template <typename T> using View2D = Kokkos::View<T **, Kokkos::HostSpace>;
+
+template <typename T> using View3D = Kokkos::View<T ***, Kokkos::HostSpace>;
+
 template <typename T> using View4D = Kokkos::View<T ****, Kokkos::HostSpace>;
 
 template <typename T> using View5D = Kokkos::View<T *****, Kokkos::HostSpace>;
 
 template <typename T> void read_array(std::ifstream &stream, View1D<T> &array);
```
Rohit-Kakodkar, line 72: Should these be a single function? We could have a function like this - 
```
template <typename ViewType>
void read_array(std::ifstream &stream, ViewType &array);
```

then have template specializations for different dimensions

fortran/meshfem3d/generate_databases/save_arrays_solver.F90
lsawade: Yes, I initially wrote a function to write an integer array instead of the boolean array. But the error was something else. I think here, I agree with the last comment, that we should assign integers to the array in terms of which element is what, and then read those integers on the CPP side into tags.

I did not do that yet to move on with the rest of the file since reading the boolean array was not the issue.

include/IO/mesh/impl/fortran/dim3/interface.hpp
lsawade: Yes, that can be done!

PR #651, opened 2025-03-17
Added test to check get_properties function

Rohit-Kakodkar:
## Description 

Implements checks for derived properties computed from base properties defined in the database. 

- [x] Implements tests to check derived properties against ground truth

## Issue Number

## Checklist

### DIFF
--- a/include/point/impl/point_data.hpp
+++ b/include/point/impl/point_data.hpp
@@ -2,6 +2,9 @@
 
 #include "datatypes/simd.hpp"
 #include "enumerations/medium.hpp"
+#include <Kokkos_Core.hpp>
+#include <string>
+#include <type_traits>
 
 #define DEFINE_POINT_VALUE(prop, index_value)                                  \
   KOKKOS_INLINE_FUNCTION value_type &prop() {                                  \
@@ -23,7 +26,7 @@ template <int N, bool UseSIMD> struct point_data {
   value_type data[N];
 
   KOKKOS_FUNCTION
-  point_data() = default;
+  point_data() {};
 
   /**
    * @brief array constructor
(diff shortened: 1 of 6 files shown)


include/point/impl/point_data.hpp
```diff
@@ -51,7 +54,7 @@ template <int N, bool UseSIMD> struct point_data {
   KOKKOS_FUNCTION
   bool operator==(const point_data<N, UseSIMD> &rhs) const {
     for (int i = 0; i < N; ++i) {
-      if (data[i] != rhs.data[i]) {
+      if (Kokkos::abs(data[i] - rhs.data[i]) > 1e-6 * Kokkos::abs(data[i])) {
```
icui, line 57: 👍

icui (approved):

include/point/impl/point_data.hpp
```diff
@@ -23,7 +26,7 @@ template <int N, bool UseSIMD> struct point_data {
   value_type data[N];
 
   KOKKOS_FUNCTION
-  point_data() = default;
+  point_data() {};
```
lsawade, line 29: Is this the preferred syntax? is there a difference? Or is it a style choice?

lsawade (approved):
This looks good to me

include/point/impl/point_data.hpp
Rohit-Kakodkar: It is not the same. `default` can mean that the constructor might be `deleted`. Here the contructor was deleted since I think `value_type data[N]` is not default constructible. By define `{}` I am explicitly telling that I can store garbage value in`data`. And its upto me to take care not to use it. This was causing an error somewhere down stream in this PR.

Rohit-Kakodkar: Dafault is always more safer.

PR #664, opened 2025-03-21
Fix CMake preset for cuda

### LINKED ISSUE #663: Kokkos_ENABLE_ATOMICS_BYPASS should be OFF for cuda presets
## Expected behavior
The optional should be off for cuda

## Files to change
`CMakePresets.json`

icui:
## Description

Set Kokkos_ENABLE_ATOMICS_BYPASS to OFF for CUDA presets.

## Issue Number

Closes #663 

## Checklist


CMakePresets.json
```diff
@@ -38,7 +38,7 @@
         "ENABLE_CUDA": "ON",
         "Kokkos_ARCH_NATIVE": "ON",
         "Kokkos_ENABLE_AGGRESSIVE_VECTORIZATION": "ON",
-        "Kokkos_ENABLE_ATOMICS_BYPASS": "ON",
+        "Kokkos_ENABLE_ATOMICS_BYPASS": "OFF",
         "Kokkos_ARCH_AMPERE80":"ON"
```
Rohit-Kakodkar, line 42: I just noticed this. I think we should not be passing `ampere` flag when compiling for cuda. We should let Kokkos determine the architecture based on which GPU it has access to. For example, This should fail when your trying to run preset on `della8` but should compile automatically for Ampere on `della-gpu` head node. 

@lsawade thoughts?

lsawade: I have not tested this preset except on della so I'm not married to anything here. I'm fine with setting enable_cuda=on and let kokkos do the rest.

So yes this is ok with me! While you are at it, maybe review the reset of the presets for errors?

lsawade (approved):

lsawade (commented):
@Rohit_Kakodkar btw, this lets you build using cmd+b in vscode using cmake tools which is neat

Rohit-Kakodkar (approved):

PR #716, opened 2025-03-28
Issue 574 - Fixing the execution from Python + VTK_CXX_BUILD=OFF works now even if VTK is found

### LINKED ISSUE #547: Add tests to compare seismograms from SH simulations
# Description

Implement seismogram tests for SH domains 
- [x] Generate seismograms from Fortran code for an SH simulation
- [x] Add SH domain traces to displacement tests
- [x] Add provenance files
- [x] Add test configuration files
- [x] Compare the seismograms using `newmark_tests.cpp`

# Acceptance criteria

- [x] Displacement tests need to pass for SH domains

### LINKED ISSUE #574: Python bindings not returning when running from Jupyter notebook.
## Expected behavior

Stop running once error occurs

## Actual (buggy) behavior

execute(config) is not returning after bindings throw error.

## Setup information:

```bash
pip install . -Ccmake.define.ENABLE_SIMD=ON -Ccmake.define.Kokkos_ENABLE_ATOMICS_BYPASS=ON -Ccmake.define.Kokkos_ARCH_NATIVE=ON -Ccmake.define.Kokkos_ENABLE_AGGRESSIVE_VECTORIZATION=ON -Ccmake.define.CMAKE_BUILD_PARALLEL_LEVEL=8
```

## Machine information:

macOS M3

lsawade:
## Description

This PR fixes the ability to interrupt the execution of the execute function from python and continue to work in the terminal. The issue at the end was that I was throwing an error in the signal_handler that is passed to the `signal(int signal, static void func(int i)` handling function from the `signal.h` C header. This caused the program to segfault. The solution is the following.

In our `periodic_tasks/check_signal.hpp` create an atomic flag:
```cpp
// Global flag for signal state - must be volatile sig_atomic_t for thread safety
volatile sig_atomic_t signal_received = 0;
```
and updated this flag through the handler:
```cpp
// Static function to handle signals
static void signal_handler(int code) {
    // Just set the flag and return
    signal_received = code;
    std::cout << "Signal " << code << " received (probably Ctrl-C)" << std::endl;
}
```
then call a super function that registers handler:
```cpp
void catch_signals() {
    // Register our handler for SIGINT (Ctrl-C)
    signal(SIGINT, signal_handler);
    
    // Currently only doing SIGINT
    //signal(SIGSEGV, signal_handler);
}
```
the periodic task function
is then:
```
void run() override {
    // Catch signals
    catch_signals();

    // Check if a signal was received
    if (signal_received) {
      // Handle the signal (e.g., throw an exception)
      std::string error_message = "Signal " + std::to_string(signal_received) +
                                   " received. Exiting...";
      // Reset the signal flag
      signal_received = 0;
      throw std::runtime_error(error_message);
    } 
  }
```

## Extra

Fixed the `VTK_CXX_BUILD=OFF` compilation which was disregarded if cmake found VTK.

## Issue Number

Closes #574 

## Checklist

### DIFF
--- a/include/periodic_tasks/check_signal.hpp
+++ b/include/periodic_tasks/check_signal.hpp
@@ -6,36 +6,6 @@
 #include <iostream>
 #include <Kokkos_Core.hpp>
 
-/**
- * @brief Catch signals
- * 
- * This function is catching the keyboard interrupt signal and other interrupt 
- * signals to parse them as exceptions to Python without crashing or hanging
- * the program. This is to avoid the program to hang via 
- * @code
- * if (PyErr_CheckSignals() != 0) {
- *     throw pybind11::error_already_set();
- * }
- * @endcode
- * if releasing the GIL in execute() function.
- */
-
-// Global flag for signal state - must be volatile sig_atomic_t for thread safety
-volatile sig_atomic_t signal_received = 0;
-
-// Static function to handle signals
-static void signal_handler(int code) {
-    // Just set the flag and return
-    signal_received = code;
-    std::cout << "Signal " << code << " received (probably Ctrl-C)" << std::endl;
-}
-
-void catch_signals() {
-    // Register our handler for SIGINT (Ctrl-C)
-    signal(SIGINT, signal_handler);
-}
-
-
 namespace specfem {
 namespace periodic_tasks {
 /**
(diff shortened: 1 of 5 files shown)


include/periodic_tasks/check_signal.hpp
```diff
@@ -10,16 +43,26 @@ namespace periodic_tasks {
...
 
   /**
    * @brief Plot the wavefield
    *
    */
   void run() override {
-    if (PyErr_CheckSignals() != 0) {
-      throw pybind11::error_already_set();
+    // Catch signals
+    catch_signals();
```
Rohit-Kakodkar, line 54: Can we put this implementation in a `.cpp` file? The reason being `catch_signals`, the volatile var, etc. are only to used within this run function. But currently anything that includes this `.hpp` file has access to those.

Rohit-Kakodkar (approved):
Minor comment.

PR #798, opened 2025-04-21
Issue 797 - Implements compile-time conditional execution of "extra physics" [poroelastic damping]

### LINKED ISSUE #797: Update Damping (/Cosserat force) calling routine
## Task description

New physics combinations need to define a has_force parameter

- [x] Add enumeration tag/attribute 
- [x] remove damping forces for function that don't have it
- [x] throw at compile time that a medium/property combination does not have "_force"

## Acceptance criteria
- [x] tests pass

lsawade:
## Description

This adds capability to have domain specific physics computations that are not required by the rest of the media. 

- Adds `default_attributes` as a base class for attributes
- default template constructor for attributes throws an error if not specialized.
- removes `damping.hpp` from `elastic` and `acoustic` media

## Issue Number

Closes #797 

## Checklist


lsawade:
Hell yeah, looks like we passing

include/enumerations/medium.hpp
```diff
@@ -61,6 +80,8 @@ class attributes<specfem::dimension::type::dim2,
   constexpr static int dimension() { return 2; }
```
icui, line 80: why is this a function?

icui (approved):

include/enumerations/medium.hpp
lsawade: @Rohit-Kakodkar ? This is not me

include/medium/compute_damping_force.hpp
```diff
@@ -0,0 +1,113 @@
...
+    const PointVelocityType &velocity, PointAccelerationType &acceleration) {
+
+  constexpr auto DimensionTag = PointPropertiesType::dimension;
+  constexpr auto MediumTag = PointPropertiesType::medium_tag;
+  constexpr auto PropertyTag = PointPropertiesType::property_tag;
+  constexpr bool has_damping_force =
+      specfem::element::attributes<DimensionTag, MediumTag>::has_damping_force;
+
+  // Compute the damping force if no implementation exists
+  if constexpr (has_damping_force) {
```
Rohit-Kakodkar, line 43: Should we replace this with std::enable_if_t above now? Since the check is happening at the construction of specfem::element::attributes<DimensionTag, MediumTag> ?

include/enumerations/medium.hpp
Rohit-Kakodkar: It could be a constexpr member variable? I did this since std::simd::size() is a function. The question is - is there a benifit of one over the other, and I havent been able to find one in my short search

icui: Perplexity says that making it a constexpr member is better (simd::size() is a function mainly for consistency and compatibility) and I think we change that.

include/medium/compute_damping_force.hpp
lsawade: Something like this:

```cpp
// For when has_damping_force is true
template<typename T, 
         std::enable_if_t<has_damping_force, int> = 0>
void impl_pre_compute_damping_force(...) {
    impl_pre_compute_damping_force(...)
}

// For when has_damping_force is false  
template<typename T,
         std::enable_if_t<!has_damping_force, int> = 0>
void impl_pre_damping_force(...) {
    return;
}
```
?

I had something like this prior just with the dispatch essentially doing the same thing. I thought we decided against that?

include/enumerations/medium.hpp
lsawade: how about:
```cpp
inline static constexpr int dimension = 2;
```

It seems as though they are identical in outcome, i.e., the compiler will optimize both equally well, but the difference in the expression without the inline is that the address of the value cannot be taken safely, an issue you don't have with the function. The `inline` key makes sure that the address can also be taken safely.

include/medium/compute_damping_force.hpp
Rohit-Kakodkar: I think it'd still nice to see an `specfem::medium::impl_compute_damping_force(...)` function that is an empty function.

would something like this work? 
```
template <typename... Args>
void specfem::medium::impl_compute_damping_force(std::false_type, const Args... /*arguments*/) {}
```

and then have function overloads with std::true_type where damping force should exist. We'd have to update the definition of `impl_compute_damping_force` to include `std::true_type` as the first argument. 

We then could also have a function that looks like this. 
```
template <typename... Args>
void specfem::medium::impl_compute_damping_force(std::true_type, const Args... /*arguments*/) {
    static_assert(always_false, "damping force function not implemented")
}
```

include/enumerations/medium.hpp
icui: In what case do you need to take the address of the value?

lsawade: This is just what I found on the web on when it's not safe to just have:
```cpp
static constexpr int dimension = 2;
```
I do not have a use case for take the address of that value. I'm just saying that without the inline, that may be unsafe.

Rohit-Kakodkar (approved):
Thanks!

include/medium/compute_damping_force.hpp
```diff
@@ -0,0 +1,170 @@
...
+    const PointVelocityType &velocity, PointAccelerationType &acceleration) {
+
+  // Extract actual tag types for the static_assert message
+  using ActualDimensionTag = typename DimensionTagType::type;
+  using ActualMediumTag = typename MediumTagType::type;
+  using ActualPropertyTag = typename PropertyTagType::type;
+
+  // The enumeration is set to true for damping force, but there is
+  // no implementation available for this dimension, medium and property
+  static_assert(specfem::utilities::always_false<ActualDimensionTag::value,
```
lsawade, line 100: @Rohit-Kakodkar we feeling ok about this?

Rohit-Kakodkar: Yes.

PR #801, opened 2025-04-22
Implements chunked domain and assembly view layouts

### LINKED ISSUE #709: Performance benchmarks without SIMD
## Task description

For Kokkos conference, I need to get performance benchmarks for versions of code without SIMD and with layout right. This would be original version I started with before making the SIMD updates 

- [ ] disable SIMD 
- [ ] update layouts to Layout::Right and update global numbering scheme
- [ ] Run benchmarks

### LINKED ISSUE #710: Get performance benchmarks w/o SIMD and without Layouts update
## Task description

Enable SIMD after implementing #709 

- [x] Enable SIMD
- [ ] ~~Run the benchmarks~~
- [x] Update `Layout::Right` to `Layout::Left`
- [x] Run the benchmarks
- [x] Update the global numbering scheme to account for SIMD computations (ispec contiguous) `issue-710-2`

### LINKED ISSUE #711: Update global indexing scheme to incorporate chunking
## Task description

Update the global numbering scheme so that `assembly_index_mapping` is chunked

- [x] Update global index mapping
- [x] Run the benchmarks

### LINKED ISSUE #713: Implement custom layouts with chunking
## Task description

Implement custom layouts to enable chunking

- [ ] Create a custom tiled layout - `specfem::kokkos::DomainLayout`
- [ ] Update `Kokkos::LayoutRight` to `specfem::kokkos::DomainLayout`
- [ ] Run the benchmarks

Rohit-Kakodkar:
## Description

This PR implements features described within #711 & #713 

- [x] Updates assembly numbering scheme such that `index_mapping` now incorporates compile time chunk sizes 
- [x] Implements domain view mapping which is a chunked kokkos view. 
- [x] Adds data containers macros which updates how data containers are created
- [x] Adds accessor methods on top of data containers
- [x] Updates existing properties and kernels containers
- [x] Updates the reader / writer for properties & kernel

## Issue Number

closes #709 
closes #710
closes #711 
closes #713 

## Performance chart 

Updated performance chart with this PR.

image

Final performance chart for GPU:

image

## Checklist


include/policies/range.hpp
```diff
@@ -166,6 +181,10 @@ struct range : Kokkos::RangePolicy<typename ParallelConfig::execution_space> {
                                             ///< type
 
   using iterator_type = specfem::iterator::range<simd>; ///< Iterator type
```
lsawade, line 183: I find the naming here confusing. How about `range_iterator_type`? just `iterator_type` makes it feels like there is choices, expecially in the documentation that does not see the implementation details.

lsawade (approved):
Looks good to me!

include/policies/range.hpp
Rohit-Kakodkar: The main reason for naming it `iterator_type` is to have a consistency. We could have pass a generic functor to either chunk policy or range policy.

```
template <typename PolicyType>
const auto lambda = [](const typename PolicyType::iterator_type::index_type index) {
   ...
}
```

PR #817, opened 2025-04-30
Added tiling for Range loops

### LINKED ISSUE #714: Implement compile time loop unrolling in range loops
## Task description

Implement compile time loop unrolling in `RangePolicy`

- [ ] Create a for loop within `Range` policy 
- [ ] Run benchmarks

Rohit-Kakodkar:
## Description

- [ ] Adds a tiling parameter for range loops
- [ ] Updates `range_iterator` and `league_iterator` methods so that starting index of the thread can be computed internall within the policies
- [ ] Adds `is_end` method to check if the iterator is out of bounds (should be only true when tile_size > 1)
- [ ] Uses `Kokkos::AUTO` when determining chunk policy configuration

## Issue Number

Closes #714 

## Checklist

### DIFF
--- a/include/policies/range.hpp
+++ b/include/policies/range.hpp
@@ -138,7 +138,7 @@ public:
   index_type operator()() const {
 #ifndef NDEBUG
     if (end_iterator) {
-      KOKKOS_ABORT("Range iterator is at the end, cannot access index.");
+      Kokkos::abort("Range iterator is at the end, cannot access index.");
     }
 #endif
     return operator()(std::integral_constant<bool, using_simd>());
--- a/include/policies/range.hpp
+++ b/include/policies/range.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "point/assembly_index.hpp"
+#include <iostream>
 #include <type_traits>
 
 namespace specfem {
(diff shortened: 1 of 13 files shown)


icui (approved):
LGTM

include/policies/range.hpp
```diff
@@ -72,7 +74,7 @@ template <typename SIMD> struct range {
 
   // range_index_type operator for simd execution
   KOKKOS_INLINE_FUNCTION
-  impl::range_index_type<true> operator()(const int i, std::true_type) const {
+  impl::range_index_type<true> operator()(std::true_type) const {
     return impl::range_index_type<true>(
         specfem::point::simd_assembly_index{ starting_index, number_points });
```
lsawade, line 79: Specific reason why this was aliased in the first place?

src/compute/assembly/helper.hpp
```diff
@@ -92,15 +92,13 @@ class helper {
...
                tile += ChunkPolicyType::chunk_size) {
-            const int starting_element_index =
-                team.league_rank() * ChunkPolicyType::tile_size + tile;
+            const auto iterator =
+                chunk_policy.league_iterator(team.league_rank(), tile);
 
-            if (starting_element_index >= nelements) {
-              break;
+            if (iterator.is_end()) {
+              return; // No elements to process in this tile
```
lsawade, line 99: A few lines below it says psv_wavefield instead of just wavefield. That must be a remnant since this is for all medium/prop combos. Update?

lsawade (approved):
Minor comments. +1 comment unrelated to PR

include/policies/range.hpp
Rohit-Kakodkar: I initially had thought we would get the tiled index using this operator here. But realized as during that implementation that its much easier to get an iterator based on the tiled index.

src/compute/assembly/helper.hpp
Rohit-Kakodkar: Fixed. Thanks!

PR #825, opened 2025-05-07
Implements new structure for examples (now benchmarks) executable directory (default ./bin) and fixes documentation

### LINKED ISSUE #479: Use CMAKE_CURRENT_SOURCE_DIRECTORY instead of CMAKE_SOURCE_DIRECTORY
If all the examples are updated like this they can be build separately, and do not need to know about the main source directory.

See the 3D example:
- [CMakeLists.txt](https://github.com/PrincetonUniversity/SPECFEMPP/blob/issue-472/examples/dim3/homogeneous_halfspace/CMakeLists.txt)

- [specfem_config.yaml]( https://github.com/PrincetonUniversity/SPECFEMPP/blob/issue-472/examples/dim3/homogeneous_halfspace/config/specfem_config.yaml.in)

### LINKED ISSUE #554: Optionally install examples and tests with pip install
Currently `pip install` creates a build directory for tests and cannot install examples. We should add an option to install tests and examples to user-defined directory and place `build/` in temporary folder.

### LINKED ISSUE #757: Building the examples in the build directory.
## Task description

This is technically quite straightforward:

- [ ] In the `cmakelists.txt` define source directory to be the `CMAKE_CURRENT_SOURCE_DIR`
- [ ] In the `cmakelists.txt` define output directory to be the `CMAKE_CURRENT_BINARY_DIR`
- [ ] Copy/Install the files that are not configured in the build directory (basically copy)

This way multiple version of SPECFEMPP can be built and examples can be run in the build folder.

***CAVEAT***:

The `build/examples` folder is already littered with cmake metabuild files making it very hard to navigate if we build the examples there.

I'm assuming that the best way to do this is to setup actuallly installation with components:
```cmake
install(TARGETS myExecutable mySharedLib myStaticLib myPlugin
        RUNTIME DESTINATION bin           COMPONENT Runtime
        LIBRARY DESTINATION lib           COMPONENT Runtime
        ARCHIVE DESTINATION lib/myproject COMPONENT Development)
```
and adapting this to
```cmake
install(FILES Par_file DESTINATION $CURRENT_BINARY_DIR COMPONENT Examples)
```

which then could be called using
```bash
cmake --install path/to/build --prefix path/to/my_examples_directory --component Examples
```

Final disclaimer, I think this can work, but I do not know.

## Acceptance criteria
- [ ] examples are built where you want them to build.

### LINKED ISSUE #829: Add Documentation for 3D mesh and assoicated specfem_config items.
## Is your feature request related to a problem?
One of the items is `database/mesh-database` and there may be more items that need to be documented.

## Describe the solution you'd like
Add to documentation
 * [ ]  What is it?
 * [ ]  When is it needed?

## Additional context

If this feature request helps with an existing issue, please link the issue here.
E.g. #<issue number>

lsawade:
## TL;DR

**Clean executable directory**

- Defaults to `root/bin`, reminiscent of fortran specfems, but can be modified with `CMAKE_INSTALL_PREFIX` and can be turnt off using `SPECFEM_INSTALL=OFF`

**`examples` -> `benchmarks/src/`**

- Now configured in a build directory compiler/architecture specific. 
- The default location is `benchmarks/build` but the location can be updated using `BENCHMARK_BUILD_DIR`.
- Updated jenkins workflows according to the new parameter, setting `BENCHMARK_BUILD_DIR=OFF`
- Each benchmark folder sets the executables in the `Snakefile` explicitly to the corresponding build to allow for easy running of multiple compilers/architecture/build_type. 

**`CMakePresets.json` updates**

- `BUILD_BENCHMARKS=ON`
- `BENCHMARK_BUILD_DIR=root/benchmarks/build/<preset>/`
- `SPECFEM_INSTALL=ON`
- `CMAKE_INSTALL_PREFIX=root/bin/<preset>/`

**Documentation Updates**

- Documentation is fully cleaned up and without errors. This means there are no excuses for implementing bad documentation. If sphinx throws a RED WARNING, FIX IT.
  - This involved a variety of fixes to the code as well as the documentation. For an implementation avoiding duplicate labels within one file see `core/specfem/point/kernel.hpp` and the corresponding `docs/sections/api/datatypes/kernels/index.rst`.
- all sections/chapters are now in `docs/sections`.
- Create Meshfem3D parameters documentation
- Updates Meshfem2D documentation to match both 3D and specfem++ documentation

## Issue Number

Would address multiple issues:

Closes #479 
Closes #554 
Closes #757 
Closes #829 

---

**In summary:**  
This PR adds optional HDF5 support for mesh/model I/O, updates the build system and documentation, and maintains backward compatibility with existing file formats.

## Original Proposal/Description

This is touching upon my mentioning that I do not like how messy the build directory is for users to "install" the software . And, the build directory really is just meant for building. 

So, here I'm proposing two things:

1. The installation of the executables (`xmeshfem2D`, `specfem2d`, etc.) into a folder `{CMAKE_INSTALL_PREFIX}/bin`, and in the future header in `{CMAKE_INSTALL_PREFIX}/include`, etc.
2. Install the examples at `-DEXAMPLE_INSTALL_DIR=/path/to/example_install_dir` or by default at `${CMAKE_INSTALL_PREFIX}`. Note that I only adapted the `homogeneous-flat-topo-example` to work with this new setup. If you all are ok with it I will move the rest.

to build

```bash
cmake --preset release
cmake --build --preset release -j6
cmake --install build/release
```

The final output will look like this if `EXAMPLE_INSTALL_DIR` is not set:

```bash
-- Install configuration: "Release"
-- Up-to-date: /Users/lsawade/specfempp/install/release/bin/xadj_seismogram
-- Up-to-date: /Users/lsawade/specfempp/install/release/bin/xmeshfem2D
-- Up-to-date: /Users/lsawade/specfempp/install/release/bin/xmeshfem3D
-- Up-to-date: /Users/lsawade/specfempp/install/release/bin/xgenerate_databases
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/Par_File
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/specfem_config.yaml
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/Snakefile
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/source.yaml
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/README.md
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/STATIONS
-- Up-to-date: /Users/lsawade/specfempp/install/release/examples/homogeneous-medium-flat-topography/topography.dat
-- Installing: /Users/lsawade/specfempp/install/release/bin/specfem2d
-- Installing: /Users/lsawade/specfempp/install/release/bin/specfem3d
```

I'm explicitly setting the install prefix in the `CMakePresets.json`
`"CMAKE_INSTALL_PREFIX": "install/release"`. Looking around online, I believe that it would not be good to set this to a random default value, because major projects are doing this the same way. HDF5 for example also requires you to set a `CMAKE_INSTALL_PREFIX` for the installation of headers/libs/executables.

Another note. To not install the dependency C++ projects (Kokkos, Gtest, yaml-cpp) I added:
`EXCLUDE_FROM_ALL` to all `FetchContents`

## Checklist

- [x] My code builds across all platforms


lsawade:
- After quite a bit of discussion last week, we arrived to the conclusion that our examples are not really serving as examples, but rather benchmarks and should be renamed as such. 
So, `examples` -> `benchmarks`.

- Moreover, since they serve as benchmarks, they should be `build`-specific benchmarks, so that gpu specific builds can be easily compared with cpu specific builds. As a result, we build the benchmarks much like documentation by using a `src` directory for the configuration files, and a `build` directory for the built benchmarks. So, benchmarks are now fully independently configurable, so that for each build you have a separate benchmarks folders. Before configuring (pre `cmake -S . -B build`) the directory looks like this:

```
benchmarks/
|-- CmakeLists.txt
|-- src/
|   |-- dim2/
|       |-- CMakeLists.txt
|           |-- example1/...
|           |-- ...
|   |-- dim3/
|   |-- ...
...
```
and if using the presets then something like this:
```
benchmarks/
|-- CmakeLists.txt
|-- build/
|   -- <preset_name>/
|       |-- dim2/
|           |-- example1/...
|               |-- specfem_config.yaml
|               |-- ...
...
|-- src/
|   |-- dim2/
|       |-- CMakeLists.txt
|           |-- example1/...
|           |-- ...
|   |-- dim3/
|   |-- ...
...
```
if working without the presets the default is to build the benchmarks directly under `benchmarks/build` so that the path for example 1 is `benchmarks/build/dim2/example1`

For a custom example build directory the user just needs to set `BUILD_BENCHMARKS` and if it is not set the default location is `benchmarks/build`

lsawade:
The second thing that was discussed is the installation in a "clean" directory. 
The standard for `cmake` projects is the following setup:
```bash
cmake -S . -B build
cmake --build build
cmake --install
```
We believe that this is misleading for our users because, we do not want people to "install" the package at the moment. So, a setup where we only
```bash
cmake -S . -B build
cmake --build build
```
is desireable. The reason why we do not want users to see the build directory is simple: it's polluted with a plethora of `cmake` build management related files, that are just confusing for the normal user. I addressed this using:
```cmake
if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
  set_property(CACHE CMAKE_INSTALL_PREFIX PROPERTY VALUE "${CMAKE_SOURCE_DIR}/bin")
endif()
```
`cmake` takes note of the fact whether `cmake` is setting the default value for the installation directory. And depending whether cmake or the user sets it we can use a default value.
And in terms of the install targets, the default will be:
```cmake
install(TARGETS specfem2d
   RUNTIME_DESTINATION .
)
```
This results in all executables being installed in `${CMAKE_SOURCE_DIR}/bin` if the `CMAKE_INSTALL_PREFIX` is not  explicitly defined by the user and directly under `${CMAKE_INSTALL_PREFIX}/` otherwise.

and for the user to set the path, the standard would be:
```bash
export PATH=path/to/specfempp_repo/bin:$PATH
```

CMakeLists.txt
```diff
@@ -7,11 +7,14 @@ option(HDF5_CXX_BUILD "Build HDF5 C++" ON)
...
 option(MPI_PARALLEL "MPI enabled" OFF)
 option(BUILD_TESTS "Tests included" OFF)
-option(BUILD_EXAMPLES "Examples included" ON)
 option(ENABLE_SIMD "Enable SIMD" OFF)
 option(ENABLE_PROFILING "Enable profiling" OFF)
 option(SPECFEMPP_BINDING_PYTHON "Enable Python binding" OFF)
+option(SPECFEM_INSTALL "Creates clean executable directory. Location Controlled by CMAKE_INSTALL_PREFIX" ON)
 option(ENABLE_DOUBLE_PRECISION "Enable double precision" OFF)
+# option(BUILD_EXAMPLES "Examples included" ON)
+option(BUILD_BENCHMARKS "Benchmarks included" ON)
```
Rohit-Kakodkar, line 16: Should the benchmarks be `OFF` by default?

Rohit-Kakodkar (approved):
Thank you! This PR is the boss, I feel you put in a insane amount of work into this and people might not see it. Really appreciate the changes. <3

icui (approved):

PR #827, opened 2025-05-09
Fixes writing of the wavefield with multiple media and adds a Python plotting script.

### LINKED ISSUE #417: Memory leak in Visualization
**Describe the bug**

Memory usage increases over time when running PNG output through VTK. 

**To Reproduce**

Run fluid-solid-bathymetry example

**Potential Solution**

I assume the window that is rendered is not properly being deleted after plotting.

lsawade:
## TL;DR

This PR updates multiple things:

1. the writer now writes the mapping that can be used for external plotting. A small bug getting the global index has been fixed.
2. the internal VTK based plotter is fully updated. What this means is:
   - The plotter now fully implements the periodic task apporach `initialize()`, `finalize()`, and `run`. This fixed the memory leak (closes #417)! 
   - the plotter also implements the same type of plotting as suggested below. The plotter keeps the old plotting using bilinear Quads by keeping the `create_biquad_grid()` method. It is manually activated however.

Side-effects of this update:
- `on_screen` plotting works now
- plotting is way faster because the unstructured grid is only created once at the start of the simulation.

### Reasoning

Plotting of the wavefield is clunky and everyone needs to create their own plotting if needs diverge just slightly. Now that @icui implemented the wavefield writing, we can use python to plot the simulation fields. However, simple interpolation of the `iglob` points or just using the corners of the elements both, frankly, suck. One does not honor topographies, and the other forgets about interior GLL points and causes artifacts in some cases (like in the currently implemented VTK plotting). 

So I thought we could potentially just plot the subquads using a poly collection to plot the subquads and update their face colors using a map from field(iglob) -> facefield(iface).

Below details, but importantly this is already working when there's only a single medium. The mapping is not written correctly, and/or I do not quite get what we are writing.

### Description

We discussed so long that I couldn't actually bring up something. So to implement a plotting tool that plots without artifacts, that is very easy to modify, I thought about the element setup: 

```
•----•-----•-----4----3
| 12 |  13 |  14 | 15 |
•----•-----•-----1----2
|  8 |   9 |  10 | 11 |
•----•-----•-----•----•
|  4 |   5 |   6 |  7 |
4----3-----•-----•----•
|  0 |   1 |   2 |  3 |
1----2-----•-----•----•
```

This is an element (GLL5) with 16 quadrilateral each with 4 vertices, and a face(color). From this you can create a simple map to average the face value from `iglob`. I marked the GLL points with numbers of vertices of the first and last subquad, and then also numbered each subquad's face.

To use this layout for plotting we need `ibool`, which can be written during the coordinate writing phase. Assuming we have the `ibool` mapping for each field, we can then get the vertices for plotting polygons and a map that maps the field `u` to `(ntotal, 4)` map which can be used to average the face of each quad.

The map you can get 
```python
for ispec, iz, ix in elements, ngllz, ngllx:
  iglob1 = mapping(ispec, iz=iz + 0, ix=ix)
  iglob2 = mapping(ispec, iz=iz + 0, ix=ix + 1)
  ...
  map[iface, :] = [iglob1, iglob2, iglob3, iglob4]

  vertices[iface, 0, 0] = x[iglob1]
  vertices[iface, 0, 1] = z[iglob1]
  ...

# Create polygon
poly = PolyCollection(vertices)

# Then later when plotting simply map the field and average in one fell swoop
ux = u[map, 0].sum(axis=1)/4.0 # for Ux in elastic_psv 
# uz = u[map, 1].sum(axis=1)/4.0 # for Uz in elastic_psv

poly.set_facecolors(cmap(norm(ux)))
```

This works actually really well and is extremely fast, because the only thing you need to do is update the face colors using ux and uz and you do not need to draw anything. 

### Bug in wavefield writer

Here, in the wavefield writer parts of the mapping are -1, meaning iglobs are -1. This does not make sense to me ([updated wavefield writer](https://github.com/PrincetonUniversity/SPECFEMPP/blob/d57d997042556bac49a76be63226cda7bbb5b1f2/include/io/wavefield/writer.tpp#L59))

Output from h5ls:
```
...
/Coordinates/acoustic/X  Dataset {125209}
/Coordinates/acoustic/Z  Dataset {125209}
/Coordinates/acoustic/mapping Dataset {7776, 5, 5}
/Coordinates/elastic_psv Group
/Coordinates/elastic_psv/X Dataset {125209}
/Coordinates/elastic_psv/Z Dataset {125209}
/Coordinates/elastic_psv/mapping Dataset {7776, 5, 5}
...
```

dumping the mapping for acoustic
[h5dump.txt](https://github.com/user-attachments/files/20124654/h5dump.txt)

And this is what I get, when I plot the raw coordinates:
```
plt.plot(acoustic_x, acoustic_z, "o", markersize=1, color="blue", label="Acoustic")
plt.plot(elastic_x, elastic_z, "o", markersize=1, color="red", label="Elastic")
```

image

and this is what I get when I plot the wavefields at timestep 'Step000600'
```
plt.scatter(acoustic_x, acoustic_z, c=np.abs(u_acoustic), s=1, cmap="Blues")
plt.scatter(elastic_x, elastic_z, c=u_tot, s=1, cmap="Grays")
```

image


Rohit-Kakodkar:
I think we need only add an if condition to solve that bug. 

```cpp
for (int i = 0; i < n_elements; i++) {
       const int ispec = element_indices(i);
       if (element_types.get_medium_tag(ispec) == __medium_tag__) {
           for (int iz = 0; iz < ngllz; iz++) {
                 for (int ix = 0; ix < ngllx; ix++) {
                     const int iglob =
                        forward.template get_iglob<false>(i, iz, ix, _medium_tag_);
                     x(iglob) = mesh.points.h_coord(0, ispec, iz, ix);
                     z(iglob) = mesh.points.h_coord(1, ispec, iz, ix);
                }
            }
        }
}

     group.createDataset("X", x).write();
     group.createDataset("Z", z).write();
});
```

examples/homogeneous-medium-flat-topography/extras/plot_wavefield.py
```diff
@@ -0,0 +1,502 @@
...
+    return fabs
+
+
+def compute_extent(plotdict, verbose=False):
+    """
+    This function computes the extent of the plot. The extent is computed
+    as the minimum and maximum x and z coordinates of the vertices.
+    """
+
+    x_min = 999999999
```
Rohit-Kakodkar, line 290: Use machine maximums here `https://numpy.org/doc/2.2/reference/generated/numpy.iinfo.html`

include/io/wavefield/writer.tpp
```diff
@@ -37,27 +41,50 @@ void specfem::io::wavefield_writer<OutputLibrary>::write(
...
-        for (int i = 0; i < n_elements; i++) {
-          const int ispec = element_indices(i);
           for (int iz = 0; iz < ngllz; iz++) {
             for (int ix = 0; ix < ngllx; ix++) {
+
+              // This is the local medium iglob
+              // see: ``count`` in specfem::compute::simulation_field<dim2, medium>
               const int iglob =
-                  forward.template get_iglob<false>(i, iz, ix, _medium_tag_);
+                  forward.template get_iglob<false>(ispec, iz, ix, _medium_tag_);
```
lsawade, line 71: the bug was here, `get_iglob` needed `ispec` and not `i`.

Rohit-Kakodkar:
Thanks for solving the memory leak. We need to discuss the PR a bit more in detail, I am not a huge fan of having separate plotting scripts in python. If the solutions implemented here using a finer element structure in VTK solves the current plotting issue, then I am in favor of just replacing the original plotting script.

lsawade:
> If the solutions implemented here using a finer element structure in VTK solves the current plotting issue, then I am in favor of just replacing the original plotting script.

By the original plotting script I assume you mean the python script? If yes, I can just remove the plotting script, but I don't think there is harm in including the mapping in the wavefield writer. There is not overhead (except creating and writing the view at the start) and the output is negligible compared to consecutive wavefield snapshots.

And yes, indeed, the new VTK setup implements the finer plotting, but if someone wants to go back to the bilinear version, that can be done using `create_biquad_grid`; just switching out one line.

lsawade:
Yes, I think this would be great either way. It would instantly enable Green function database storage for the full domain. So I'd be for it without a doubt! 
Without the need for mapping. I'm all for it, but thought a separate PR maybe better for it.

Rohit-Kakodkar:
Sounds good. This PR looks good to me.

Rohit-Kakodkar:
Is this PR ready to be merged? @icui waiting for review here.

icui (approved):

PR #833, opened 2025-05-14
Point properties/kernels to remove manual indexing

Rohit-Kakodkar:
## Description

- [ ] Adds data container macros to generate properties/kernels classes
- [ ] Updated point properties/kernels

## Issue Number

## Checklist


include/point/properties.hpp
```diff
@@ -1,115 +1,90 @@
...
-  DEFINE_POINT_VALUE(c15, 2) ///< @f$ c_{15} @f$
-  DEFINE_POINT_VALUE(c33, 3) ///< @f$ c_{33} @f$
-  DEFINE_POINT_VALUE(c35, 4) ///< @f$ c_{35} @f$
-  DEFINE_POINT_VALUE(c55, 5) ///< @f$ c_{55} @f$
-  DEFINE_POINT_VALUE(c12, 6) ///< @f$ c_{12} @f$
-  DEFINE_POINT_VALUE(c23, 7) ///< @f$ c_{23} @f$
-  DEFINE_POINT_VALUE(c25, 8) ///< @f$ c_{25} @f$
-  DEFINE_POINT_VALUE(rho, 9) ///< Density @f$ \rho @f$
-  ///@}
+  POINT_CONTAINER(c11, c13, c15, c33, c35, c55, c12, c23, c25, rho)
```
lsawade, line 87: where is that beautiful documentation?

include/point/kernels.hpp
```diff
@@ -1,299 +1,139 @@
...
-   */
-  ///@{
-  DEFINE_POINT_VALUE(rho, 0)   ///< \f$ K_{\rho} \f$
-  DEFINE_POINT_VALUE(mu, 1)    ///< \f$ K_{\mu} \f$
-  DEFINE_POINT_VALUE(kappa, 2) ///< \f$ K_{\kappa} \f$
-  DEFINE_POINT_VALUE(rhop, 3)  ///< \f$ K_{\rho'} \f$
-  DEFINE_POINT_VALUE(alpha, 4) ///< \f$ K_{\alpha} \f$
-  DEFINE_POINT_VALUE(beta, 5)  ///< \f$ K_{\beta} \f$
-  ///@}
+  POINT_CONTAINER(rho, mu, kappa, rhop, alpha, beta)
```
lsawade, line 25: same here

include/point/impl/point_container.hpp
```diff
@@ -0,0 +1,200 @@
...
+namespace specfem {
+namespace point {
+
+namespace impl {
+namespace properties {
+
+template <specfem::dimension::type Dimension,
+          specfem::element::medium_tag MediumTag,
+          specfem::element::property_tag PropertyTag, bool UseSIMD>
+struct point_traits {
```
lsawade, line 161: I though `point_traits` was going to be the same struct for all points containing both
```cpp
constexpr static bool is_point_properties = true;
constexpr static bool is_point_kernels = false;
```
similar to the implementation of [ImplFieldTraits](https://github.com/PrincetonUniversity/SPECFEMPP/blob/8e45951fa1b7037a8c4e4a95f52f0796ef98062b/include/point/field.hpp#L80)

I think if we do the traits containing `is_xy` then it should be consistent across the datatypes

Rohit-Kakodkar: Yes, you are right. I need name this differently. `point_traits` should be reserved for general point types

include/point/properties.hpp
Rohit-Kakodkar: I was gonna do it when I resolved merge conflicts with your branch. But I should do it now.

icui (approved):

lsawade (approved):

PR #1030, opened 2025-07-17
Fixes allocation of memory for boundary values to when needed

Rohit-Kakodkar:
## Description

- [x] Allocate memory only when writer is initialized
- [x] Update `load` and `store` functions
- [x] Updates function definitions so that writer information can be passed to assembly

@maxlchien so with this PR you should not see a memory allocation in boundary values as long as you dont specify a `writer` node within forward simulations in `specfem_config.yaml`

Update since review: 
- [x] Check if the the writer is initialized with the intention on running adjoint simulations using `for_adjoint_simulations` parameter
- [x] Allocate memory only when needed
- [x] Update wavefield reader and writer
 
## Issue Number

## Checklist


lsawade (approved):
LGTM, I checked the jenkins tests, it looks like only displacement BC tests are failing with SEGFAULTs, so I assume the allocation isn't done and the boundary values are accessed.

icui (approved):

lsawade (approved):
It was an accident...

PR #1081, opened 2025-08-04
Use references for 'const std::string' function arguments.

bangerth:
Copying `std::string` objects is pretty expensive. When a function (or constructor) takes a `const std::string` argument, the `const` indicates that it has no interest in modifying the string, and so it should take the argument by `const` reference instead to avoid the copy.

### DIFF
--- a/core/specfem/receivers/dim2/receiver.hpp
+++ b/core/specfem/receivers/dim2/receiver.hpp
@@ -31,7 +31,7 @@ public:
    * @param z Z coordinate of the station
    * @param angle Angle of the station
    */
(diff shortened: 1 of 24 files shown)


buildbot-princeton:

buildbot-princeton:

lsawade:
Hi @bangerth, we are using `devel` as our base development branch (we didn't get to the github workflow in the workshop). Could you either update the branch, or let us know if you want us to implement the changes?

Rohit-Kakodkar (approved):
Thanks for the fix!

Rohit-Kakodkar:
Test this please

PR #1069, opened 2025-08-01
Adds adjacency map computation to the mesher

Rohit-Kakodkar:
## Description

- [x] Adds adjacency map computation
- [x] Writes adjacency map to database
- [x] Removes older implementation

## Issue Number

If there is an issue created for these changes, link it here

## Checklist

Please make sure to check developer documentation on specfem docs.

- [ ] I ran the code through pre-commit to check style
- [ ] **THE DOCUMENTATION BUILDS WITHOUT WARNINGS/ERRORS**
- [ ] I have added labels to the PR (see right hand side of the PR page)
- [ ] My code passes all the integration tests
- [ ] I have added sufficient unittests to test my changes
- [ ] I have added/updated documentation for the changes I am proposing
- [ ] I have updated CMakeLists to ensure my code builds
- [ ] My code builds across all platforms

Issue #517, opened 2025-02-26
Discontinous galerkin


Rohit-Kakodkar:
Base branch for discontinous galerkin implementation

- [ ] Iteration policy, chunk edge view

@int-ptr-ptr update this issue title as you keep issuing PRs

PR #1089, opened 2025-08-08
dG - Edge Containers

int-ptr-ptr:
## Description

A minor feature for the broader dG project (#517).
This PR branches from #1046, so that PR should be merged first.

While the adjacency graph (#1069) is in progress, we are implementing the nonconforming interface containers. The design of these containers is / will be based on modular inheritance, similar to the draft [here](https://github.com/int-ptr-ptr/SPECFEMPP/blob/dg-phase2/workflow_int_ptr_ptr/lci_plan.md).

Before the first dG kernel can be written, the nonconforming-coupled-interface containers need to incorporate edge index mappings, mortar transfers, and normal vector storage. Each of these features will be given a module (a struct to inherit) in order to keep code DRY between schemes. This PR provides the edge index mappings through two modules:
- `single_edge_container`
  - has `index_mapping` for `iedge -> ispec` and `edge_type` for `iedge -> edge type`
  - Used for when a medium interfaces itself, and it makes more sense to use a single edge list (i.e. fully nonconforming mesh)
  - The interface component will, for each interface element, point to two edges in the container.
- `double_edge_container`
  - `index_mapping`s for each of two media. This looks like the conforming `coupled_interface` already in SPECFEM.
  - Used for when there are two separate media or it makes sense to have two separate edge lists.
  - The interface component will, for each interface element, point to an edge on each medium in the container.

We also introduce the backbone for the module unit tests. For now, interfaces will be provided in a separate file until the adjacency graph is complete. Containers, isolated from the `assembly` struct, are declared and initialized to match this file.

For now, the only part of the test that is populated verifies that the containers work with the current `specfem::policy::element_edge` policy.

## Issue Number

## Checklist


int-ptr-ptr:
> @int-ptr-ptr, is this good to review?

This PR depends on #1046, so it's probably better to wait for that to be merged, first. I'll request the reviews afterwards.

PR #1091, opened 2025-08-08
Enabled static C++ HDF5 installation, fixed policies

### LINKED ISSUE #1077: CMake warning about file suffixes.
I get the following warning when running `cmake ..` in a build directory within the source directory, using CMake 3.28:
```
-- Configuring done (102.7s)
CMake Warning (dev) at fortran/meshfem3d/generate_databases/CMakeLists.txt:92 (add_library):
  Policy CMP0115 is not set: Source file extensions must be explicit.  Run
  "cmake --help-policy CMP0115" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  File:

    /home/bangerth/p/deal.II/1/projects/SPECFEMPP/fortran/meshfem3d/generate_databases/model_sep_nompi.f90
This warning is for project developers.  Use -Wno-dev to suppress it.
```
I must admit that I'm not entirely sure what that is supposed to mean. The file listed exists and, at least to me, `.f90` is an explicit file extension. The only thing I can imagine is that CMake has not realized that it's dealing with a Fortran90 file.

### LINKED ISSUE #1082: cmake hdf5 option not working
## Task description

Put HDF5_FOUND in HDF5_CXX_BUILD block

## Acceptance criteria
- [ ] build builds, without hdf5, even though it is loaded or present in /include

lsawade:
## Description

- Enables local static installation of HDF5. It does **not** significantly increase build time.
- Fixes cmake policies mentioned by @bangerth
- Implements GHA workflow to test installation of external IO libraries.

## Issue Number

Closes #1077 . Closes #1082 

## Checklist


cmake/kokkos.cmake
```diff
@@ -1,3 +1,6 @@
+# To work for CMake 3.28 and below
+cmake_policy(SET CMP0169 OLD)
```
bangerth, line 2: This only papers over the issue. Policies are always assumed to deprecate stuff, and setting the policy to `OLD` will simply make the issue re-appear at some point in the future with a sufficiently new CMake.

lsawade: Yes, indeed, but to solve all the issues i'm setting with the policies i need to substantially increase the minimum cmake version, which i do not want to do at this time. E.g. for the boost policy, the simplest thing would be to change how we find it, but that would only be allowed from certain cmake versions. For now, except for the policies we do not need a higher version of cmake.

Is this not the way i should be solving this in that case?

cmake/hdf5.cmake
```diff
@@ -2,22 +2,79 @@
...
+
+    # If not found or force install is enabled, fetch and build HDF5 from source
+    if((NOT HDF5_FOUND) OR (SPECFEM_ENABLE_HDF5_FORCE_INSTALL))
 
-    if (NOT ${HDF5_FOUND})
-        message(STATUS "HDF5 not found. Building without HDF5.")
-        set(SPECFEM_ENABLE_HDF5 OFF CACHE BOOL "Disable HDF5 support" FORCE)
-    else()
+        # Disable unity build for here
+        set(SAVE_UNITY_BUILD ${CMAKE_UNITY_BUILD})
```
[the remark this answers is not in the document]
Rohit-Kakodkar: @lsawade see this

Rohit-Kakodkar (approved):

icui (approved):

lsawade:
Alrighty, the build should be passing now, but the solution to use our, to adios2 external `yaml-cpp`. 

This is only an intermediate solution because fetchcontent_populate is going to be deprecated, I have made an issue on the adios2 side ([ADIOS2 - yaml-cpp issue](https://github.com/ornladios/ADIOS2/issues/4612)) to see whether they have an idea on a work around that does not require a already installed yaml library.

Issue #1147, opened 2025-08-27
gmesh external mesher


int-ptr-ptr:
## Task description

- [x] script takes in the bathymetry and associated mesh values
- [x] Read in topo file into gmesh

## Task Acceptance 

- [x] Fluid Solid non-confirming interface equivalent bathymetry as internal mesher

PR #1148, opened 2025-08-27
gmsh: topography file to external mesh files

int-ptr-ptr:
## Description

Adds a script `gmshlayerbuilder` to the `scripts` directory. Since `argparse` is used, a help message is provided using the `-h` flag.

A simple example topography file has been added into the script, allowing a quick demo using
```sh
python scripts/gmshlayerbuilder scripts/gmshlayerbuilder/test_topo.dat <output_folder>
```

As of now, since this script is meant to be called, the body of the logic is in an underscored submodule.

## Issue Number

#1147 

## Checklist


Rohit-Kakodkar (approved):
This looks great! Could you add a README.md that'd let you generate the required files using the topography file

scripts/gmshlayerbuilder/_gmshlayerbuilder/dim2/binary_detect_N3.py
```diff
@@ -0,0 +1,394 @@
+import numpy as np
```
lsawade, line 1: Is this file used?

lsawade (changes requested):
Some comments. I walked through the execution, and it looks pretty good.

Some general comments, and a logic tree that I think (corrected by you) should probably be put in some form of documentation ()

```md

toporeader
- reads LerpLayerBoundary
  - contains points and implements build_layer
- creates layers using interfaces
- assigns everything to builder

builder
- creates model
  - loops over boundaries to create lines
    - LerpLayerBoundary.build_layer(...)
      - LayerBoundary.BuildResult()
        - creates the gmesh polyline from the vertices 
  - loops over and layer boundaries
    - LerpLayerBoundary.initialize_curve_copy() 
      - checks "automatically" whether the layers are conforming
      - if yes creates copy of above boundary
      - if no creates no copy
    - loops over layers, and pairs of boundaries 
      - Layer.generate_layer(...)
        - generates the actual layers medium blocks in gmesh
  - generates mesh from gmesh building blocks
  - creates model from meshed surfaces

- Export the files
```

In general, I think you are overusing `__all__` resulting in a very hard to navigate code. Other than that this looks really good to me.

I am confused about the lagrange intersection finding, that is, model.union? Is that not taken care of by gmesh if you define different layers with different discretizations?

Note that I have not looked deeply into gmesh, but I assumed that was the case?

scripts/gmshlayerbuilder/_gmshlayerbuilder/dim2/binary_detect_N3.py
int-ptr-ptr: `quadratic_beziers_intersect` is called in nonconforming interface finding.

lsawade: Yes, figured this out later, hence comment:
> I am confused about the lagrange intersection finding, that is, model.union? Is that not taken care of by gmesh if you define different layers with different discretizations? Or does Gmesh not give adjacent edges. I know rohit had to write his own, less-involved snippet for cubit. 

You have nothing here that calls model.union yet though correct? and model.union uses this.

lsawade (changes requested):
Much better! Thank you for these updates. As discussed earlier let's put the exporter into a separate place. Tbh, I like you __main__.py setup. 

so if we have
- `export_cubit/__main__.py`
- `export_gmsh/__main__.py`
- `create_gmsh_from_topography_file/__main__.py`

This would require scripts to have `__init__.py` which I am personally not a fan of. Maybe you have an idea?

int-ptr-ptr:
> This would require scripts to have __init__.py which I am personally not a fan of. Maybe you have an idea?

If we lock people into using the root `pyproject.toml` for all of the scripts, you could install (i.e. add to module path) the modules like `export_gmsh` on a `uv sync` or `pip install .` The benefits to that I immediately see are
- no `sys.path` manipulation needed
- `[project.scripts]` command-line entry points anywhere inside the module, so `__main__.py` is not required.
- Since the python environment is already aware of the modules, my `_gmshlayerbuilder` subdirectory is unnecessary.
- examples (and user) have an absolute reference / alias to scripts and modules via python virtual environment
- straight-forward script / module testing, as is typical for `pyproject`s. Especially for `gmshlayerbuilder`, the modules are large enough to warrant unit testing.

For now, I would want to push this script through -- I already did a bit of reorganizing, and I think it would be beneficial to have the runnable done now rather than to keep having a back and forth on how best to organize the files. This would also be a good conversation to include Congyue on, as well.

lsawade (approved):
LGTM!

Rohit-Kakodkar (approved):
I agree with Lucas. We should split the exported from this implementation into a separate script. We can do it later, but probably before the merge of DG implementation within devel (or next release if we're ready with DG by then). 

For now this looks good. Great work!

PR #1186, opened 2025-09-09
Compute Reparameterization for Nonconforming Interfaces

### LINKED ISSUE #1173: NCI mortar reparameterization
## Task description

Write a list of things that need to be completed to achieve the goal of this issue.

Create namespace `specfem::assembly::nonconforming_interfaces` with `dim2` function:
- [x] `compute_intersection(adjacency_graph_edge, mortar_quadrature) -> vector<pair<source_local_coord, target_local_coord>>` - builds a reparameterization on the intersection between two elements, described by an edge on the adjacency graph. `ret_val[i]` is the pair of edge local coordinates of the two elements corresponding to quadrature point `i` of the mortar.

Revive `unit-tests/nonconforming` with tests:
- [x] verify `compute_intersection` gives intersections that are a disjoint union of the entire interface.
- [x] verify the local coordinates for each mortar quadrature point map to the same global coordinates.

## Acceptance criteria

- [x] The test passes

int-ptr-ptr:
## Description

(See issue #1173)

Adds a way to compute the local coordinate reparameterization across an intersection

## Issue Number

solves #1173

## Checklist


buildbot-princeton:

Rohit-Kakodkar (commented):
I think apart from the changes we just talked about for tests, this look good. With minor comments.

core/specfem/assembly/nonconforming_interfaces/dim2/impl/compute_intersection.cpp
```diff
@@ -0,0 +1,166 @@
...
+    const Kokkos::View<
+        specfem::point::global_coordinates<specfem::dimension::type::dim2> *>
+        &element1,
+    const Kokkos::View<
+        specfem::point::global_coordinates<specfem::dimension::type::dim2> *>
+        &element2,
+    const specfem::mesh_entity::type &edge1,
+    const specfem::mesh_entity::type &edge2,
+    const Kokkos::View<type_real *> &mortar_quadrature) {
+  constexpr type_real eps = 1e-5;
```
lsawade, line 52: what if type_real is double precision?

core/specfem/assembly/nonconforming_interfaces/dim2/impl/compute_intersection.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  // recover the bounds of the intersection in local coordinates:
+  // get_local_edge_coordinate returns values within [-1,1], so an edge's own
+  // bounds are already factored in.
+  type_real inter_min_on_1 = std::min(p2lo_on_1, p2hi_on_1);
+  type_real inter_max_on_1 = std::max(p2lo_on_1, p2hi_on_1);
+  type_real inter_min_on_2 = std::min(p1lo_on_2, p1hi_on_2);
+  type_real inter_max_on_2 = std::max(p1lo_on_2, p1hi_on_2);
+
+  if (inter_min_on_1 > inter_max_on_1 - eps ||
```
lsawade, line 88: use `include/utilities/utilities.hpp::is_close(...)` ?

lsawade (changes requested):
Two things, otherwise it looks good to me.

1. Throughout the testing suites, we are using Google tests EXPECT statements where we `<< ExpectedGot(expected, got)` to print unexpected outputs. I understand that you want better printing, but then maybe update `ExpectedGot(...)` struct from `test_macros.hpp`
  ```cpp
  EXPECT_TRUE(specfem::utilities::is_close(expected, got)) 
    <<  ExpectedGot(expected, got) 
    << "Extra info: that you want to print in case of failure e.g: "
    << i 
    << std::endl;
  ```
  to improve printing for all tests this was implemented to take care of `simd` types, and would reduce the code verbosity by a lot and you don't have to manually assign when the test fails. GoogleTest then also takes care of telling you where the error is exactly occurring.
2. Along the same lines you spent quite some time to create a fail-safe test for the edge in case elements are not aligned. That should be tested with a `EXPECT_THROW()` test.

core/specfem/assembly/nonconforming_interfaces/dim2/impl/compute_intersection.cpp
int-ptr-ptr: For ~~this and~~ (1) above, can we push that back to a later PR? mortar transfer computation is dependent on this, so I want to get this pushed through so that I can implement that.

I've [added a task to triage](https://github.com/orgs/PrincetonUniversity/projects/57/views/2?pane=issue&itemId=129220787) that would hopefully address your concerns, since this pattern of needing to check multiple values together will be useful for multiple things.

Edit {~~this and~~}: Realizing that this was not the file I was thinking about. See comment on `eps`.

core/specfem/assembly/nonconforming_interfaces/dim2/impl/compute_intersection.cpp
int-ptr-ptr: Here, `eps` is only used to add a fuzziness in whether intervals overlap, so it's more geometric that numerical.

This is currently a stop-gap solution for an issue I realized:

> I do not know how we would want to handle intersections that are empty or near empty. `cubit` or `gmsh` may present an intersection between `(ispec, iedge) <-> (jspec, jedge)` that spans a tiny interval (or is disconnected by a tiny amount) that we would want to ignore `specfem`-side. I don't think it's worth throwing an error over, since I don't believe it to be an issue with user meshing.
> Best case, I think, would be to prune these intersections, which would require a bit more complicated layout. For now, I would expect it to be fine to leave them in (they won't ruin the numerics, since the integrals are vanishingly small), so I added a small tolerance.

lsawade (approved):
LGTM

icui (approved):

core/specfem/assembly/nonconforming_interfaces/dim2/compute_intersection.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "quadrature/interface.hpp"
+#include "quadrature/quadrature.hpp"
+#include "specfem/assembly.hpp"
+
+#include <vector>
+
+namespace specfem::assembly {
+namespace nonconforming_interfaces {
```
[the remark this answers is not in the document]
Rohit-Kakodkar: This should be `nonconforming_interfaces_impl` right? I am assuming this will be only used to build the nonconforming interfaces data container. You'd want to call your data container `specfem::assembly::nonconforming_interfaces`.

include/mesh/dim2/adjacency_graph/adjacency_graph.hpp
```diff
@@ -61,12 +60,13 @@ template <specfem::dimension::type Dimension> struct adjacency_graph {
    * - No vertex properties
    * - EdgeProperties for edge data
    */
-  using Graph =
+  using GraphType =
```
Rohit-Kakodkar: This can be in private again, since you templated the function that used it before.

Rohit-Kakodkar (changes requested):
Couple of minor changes. Otherwise this looks good.

Rohit-Kakodkar (approved):
Awesome! Thanks!

Rohit-Kakodkar:
ok to test

Rohit-Kakodkar:
@int-ptr-ptr some tests failed.

int-ptr-ptr:
I'm realizing that after the `EdgeType` templating, the `.cpp` should have gone to `.tpp`, but it looks like whatever CMake is doing in `core/specfem` doesn't care and is not generating the object files?

Oh! The current tests are assembly agnostic! No error is thrown, since no functions passing `EdgeType` are called. I'll change that in a future PR (since I don't want to handle the merge conflict from a file relocation ;P).

PR #1202, opened 2025-09-12
Adds a skeleton for non conforming kernel implementation [DO NOT MERGE]

### LINKED ISSUE #1153: Implement kokkos kernel for non-conforming interfaces
## Task description

This is an outline of the kernel that Kentaro can populate. 

- [ ] Write outline kernel for computation of non-conforming coupled interfaces

Rohit-Kakodkar:
## Description

@int-ptr-ptr I added a skeleton for the non-conforming kernel, which we can fill out after we have the containers. Can you see if this looks right? 

## Issue Number

## Checklist


int-ptr-ptr (approved):
This looks right, but I do wonder about the future, where coupled interfaces have shared computations (for example, jump operators), so it would be faster to accelerate both *self* and *other* at the same time.

Since interfaces are $O(N)$ with the length scale resolution, this won't be an issue for Ivy's problem (we can swallow the unoptimized code for now), but it would be something to keep in mind.

lsawade:
Addressed in PR #1265

Issue #1170, opened 2025-09-05
Implement 3D algorithms


Rohit-Kakodkar:
## Task description

Write a list of things that need to be completed to achieve the goal of this issue.

- [ ] Implement 3D gradient algorithm 
- [ ] Implement 3D divergence algorithm

## Acceptance criteria
- [ ] Find ways to best test the algorithms

icui:
<img width="500" height="400" alt="Image" src="https://github.com/user-attachments/assets/b9010a8b-081d-4767-9982-d4a666420c10" />

PR #1224, opened 2025-09-18
1170 - [Proposal] Move element-level divergence calculation to chunk_element

icui:
## Description

I think by definition `divergence` and `gradient` calculation should be a method of `Tensor`, the function for the calculation of the entire domain remains the same `algorithms::divergence`. When applied to `gradient` in the future this will also eliminate the repetitive code for one and two field calculations.

## Issue Number

Adds to #1170

## Checklist

### DIFF
--- a/include/datatypes/chunk_element_view.hpp
+++ b/include/datatypes/chunk_element_view.hpp
@@ -518,9 +518,10 @@ struct TensorChunkViewType<T, specfem::dimension::type::dim2, NumberOfElements,
       typename specfem::point::index<dimension_tag, UseSIMD>; ///< index type
                                                               ///< for accessing
                                                               ///< at GLL level
-  using point_view_type = TensorPointViewType<T, Components, NumberOfDimensions,
-                                              UseSIMD>; ///< Point view type for
-                                                        ///< component access
+  using point_view_type =
+      TensorPointViewType<T, Components, NumberOfDimensions,
+                          UseSIMD>;           ///< Tensor Point view type for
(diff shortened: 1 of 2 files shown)


include/datatypes/chunk_element_view.hpp
```diff
@@ -607,6 +608,40 @@ struct TensorChunkViewType<T, specfem::dimension::type::dim2, NumberOfElements,
...
+    }
+
+    VectorPointViewType<T, Components, UseSIMD> result;
+    for (int icomp = 0; icomp < components; ++icomp) {
+      result(icomp) = weights(iz) * temp1l[icomp] + weights(ix) * temp2l[icomp];
+    }
+
+    return result;
+  }
 };
```
Rohit-Kakodkar, line 645: This is not the divergence, though. It only computes the value of divergence at a single GLL point.

icui: Are you worried about the naming or prefer the current organization?

Rohit-Kakodkar: Okay, I am wrong, this is the same equation as Komatitsch in 1999. Having this in Tensor puts restrictions on how divergence can be called.  

For example, we would still use the "same" divergence algorithm on a 2D surface as a 3D element. The only thing that changes between the 2 is the quadrature scheme. I like your idea, but we should discuss it more.

lsawade (commented):
You're going to make updates to this, right, @icui ?

PR #1301, opened 2025-10-17
Accessor Draft for Nonconforming Interfaces

int-ptr-ptr:
## Description

- moves field `load_on_host` functions to `data_access`
- combined conforming and nonconforming pre-impl accessors for `load_on_host` and `load_on_device`
- `edge_normal` -> `intersection_normal`. Stores on interface quadrature points.
- updated test for new naming / organization.

Additionally, the part I would like feedback on is this:
- draft accessor selection (see `load_access_compatibility.hpp`, which creates flags for "does this container have this field?")
- `CheckCompatibility2` replaces the point check with an accessor equivalence. Should this replace the original `CheckCompatibility`?

## Issue Number

## Checklist


core/specfem/chunk_edge/nonconforming_coupled_interface/transfer.hpp
```diff
@@ -0,0 +1,242 @@
...
+ * @tparam BoundaryTag Boundary condition applied to the interface
+ * @tparam MemorySpace Memory space for data storage.
+ * @tparam MemoryTraits Memory traits for data storage.
+ * @tparam UseSIMD Flag to indicate if SIMD should be used.
+ */
+template <bool IsSelf, int NumberElements, int NQuadElement, int NQuadInterface,
+          specfem::interface::interface_tag InterfaceTag,
+          specfem::element::boundary_tag BoundaryTag, typename MemorySpace,
+          typename MemoryTraits, bool UseSIMD>
+struct nonconforming_transfer_function<
```
lsawade: SHould this be templated with an enumeration being the "transfer type"

int-ptr-ptr: Not sure what you mean. The transfer function is a linear map from *the span of* edge shape functions to *the span of* interface shape functions, so it's always represented as a $\dim(\verb|edge|) \times \dim(\verb|intersection|)$ matrix. `self` and `coupled` are already specified by the `bool` at the top.

Edit: Italicized text: "the span of". That is to say the scalar fields in the corresponding finite dimensional subspaces of $H^1$. If you're talking about transferring vectors or tensors, they get mapped using the same linear function.

core/specfem/data_access/check_compatibility.hpp
```diff
@@ -26,6 +26,26 @@ struct CheckCompatibility {
...
+      ContainerType::data_class == AccessorType::data_class;
+
+public:
+  static constexpr bool value = check_dimension && check_data_class;
+
+  static_assert(check_dimension, "Dimension tags do not match");
+  static_assert(check_data_class, "Data classes do not match");
+  static_assert(specfem::data_access::is_container<ContainerType>::value,
+                "ContainerType is not a container");
+  static_assert(AccessorType::accessor_type == IndexType::accessor_type,
```
Rohit-Kakodkar, line 24: Should the 2 functions be merged?

core/specfem/assembly/nonconforming_interfaces/dim2/impl/nonconforming_interface.hpp
```diff
@@ -109,23 +112,142 @@ struct interface_container<specfem::dimension::type::dim2, InterfaceTag,
...
+                  specfem::assembly::coupled_interfaces_impl::
+                      stores_intersection_normal<EdgeType>::value) {
+      Kokkos::parallel_for(
+          Kokkos::TeamThreadRange(team, num_edges * EdgeType::n_quad_interface),
+          [&](const int &ichunkmortar) {
+            const int imortar = ichunkmortar % EdgeType::n_quad_interface;
+            const int iedge = ichunkmortar / EdgeType::n_quad_interface;
+
+            const int &local_slot = iedge;
+            const int &container_slot = offset + iedge;
```
Rohit-Kakodkar, line 174: I see. There should technically be `get_local_index` for an iterator. I dont think its implemented for `edge` or `intersection` iterators. I can create an issue and update this later.

Rohit-Kakodkar (approved):
This looks good for now. The access functions and chunk edge types for non-conforming interfaces can be simplified later; I’ll approve this for now and refactor those aspects after their usage is defined in future PRs.

core/specfem/data_access/check_compatibility.hpp
int-ptr-ptr: as in the description
> - `CheckCompatibility2` replaces the point check with an accessor equivalence. Should this replace the original `CheckCompatibility`?

If that's what you want, then I can merge them; it's just that I haven't checked if the things that use `CheckCompatibility` require separate point verification.

core/specfem/chunk_edge/impl/nonconforming_transfer_function.hpp
```diff
@@ -0,0 +1,242 @@
...
+ * @tparam MemoryTraits Memory traits for data storage.
+ * @tparam UseSIMD Flag to indicate if SIMD should be used.
+ */
+template <bool IsSelf, int NumberElements, int NQuadElement, int NQuadInterface,
+          specfem::dimension::type DimensionTag,
+          specfem::connections::type ConnectionTag,
+          specfem::interface::interface_tag InterfaceTag,
+          specfem::element::boundary_tag BoundaryTag, typename MemorySpace,
+          typename MemoryTraits, bool UseSIMD>
+struct nonconforming_transfer_function;
```
icui, line 37: If this struct is public in `chunk_edge` namespace, the file should also be placed in `chunk_edge` instead of `chunk_edge/nonconfirming_coupled_interface`.

core/specfem/chunk_edge/nonconforming_coupled_interface.hpp
```diff
@@ -0,0 +1,172 @@
...
+          typename MemoryTraits, bool UseSIMD>
+struct nonconforming_coupled_interface<
+    NumberElements, NQuadElement, NQuadInterface,
+    specfem::dimension::type::dim2, specfem::connections::type::nonconforming,
+    InterfaceTag, BoundaryTag, MemorySpace, MemoryTraits, UseSIMD>
+    : public specfem::data_access::Accessor<
+          specfem::data_access::AccessorType::chunk_edge,
+          specfem::data_access::DataClassType::coupled_interface,
+          specfem::dimension::type::dim2, false>,
+      public specfem::chunk_edge::impl::nonconforming_transfer_function<
```
icui, line 64: should this be `specfem::chunk_edge::impl::nonconforming_transfer_function` or `specfem::chunk_edge::nonconforming_transfer_function`?

icui (changes requested):

core/specfem/data_access/check_compatibility.hpp
Rohit-Kakodkar: I see. Yes, the above has a bug. The compatibility should be valid for all indices and not just point.

core/specfem/chunk_edge/nonconforming_coupled_interface.hpp
int-ptr-ptr: `impl::nonconforming_transfer_function` does not inherit `specfem::data_access::Accessor`, which is important because of the multiple inheritance in `nonconforming_coupled_interface`. It needs to be `impl`

core/specfem/chunk_edge/impl/nonconforming_transfer_function.hpp
Rohit-Kakodkar: @int-ptr-ptr I think we need some restructuring here. Can you move everything within the impl namespace to `core/specfem/chunk_edge/impl/nonconforming_transfer_function.hpp`? Create two files within the chunk_edge folder: one for nonconforming_coupled_interface.hpp (which already exists) and another for nonconforming_transfer_function.hpp (for specfem::chunk_edge::nonconforming_transfer_function). 

Our standard is - the namespace should define where the file lies.

core/specfem/chunk_edge/nonconforming_coupled_interface/transfer.hpp
lsawade: I realize my question did not make sense the way phrased it. Neither is this the place to address it.

My actual question should have been: What was the plan for the flux scheme again? Is the plan to address different flux schemes through `include/medium/compute_coupling`?

lsawade (approved):
@Rohit-Kakodkar 's comments should be addressed, but other than that I'm ok with this!

icui (approved):

Issue #1264, opened 2025-10-02
Update /include/element/quadrature and assembly/mesh load/store quadrature


lsawade:
## Task description

updating the quadrature holder is require for the wavefield computation

- [ ] Update the quadrature holder to compute the quadrature on 3D array 
- [ ] Update assembly/mesh) load/store quadrature functions

## Acceptance criteria
- [ ] The tests for the new function `bar()` pass

PR #1309, opened 2025-10-20
1264 Part III - Create md range iterator for rank 2 and 3

icui:
## Description

Benchmark shows that `Kokkos::TeamThreadMDRange` significantly downgrades the performance, so a custom iterator for quadrature is necessary. 

## Issue Number

Adds to #1264

## Checklist


include/execution/chunked_domain_iterator.hpp
```diff
@@ -242,8 +242,8 @@ class ChunkElementIterator
                             : simd_size;
     int ispec = indices(ielement);
     int xz = i / num_elements;
-    const int iz = xz / element_grid.ngllz;
-    const int ix = xz % element_grid.ngllz;
+    const int iz = xz / element_grid.ngllx;
```
Rohit-Kakodkar, line 245: Why did you change the ordering here?

include/execution/team_thread_md_range_iterator.hpp
```diff
@@ -0,0 +1,92 @@
...
+                            ///< this iterator.
+
+  using execution_space =
+      typename base_type::execution_space; ///< Execution space type.
+
+  KOKKOS_INLINE_FUNCTION const index_type operator()(const int &i) const {
+    const int i1 = i / N2;
+    const int i2 = i % N2;
+    return index_type(i1, i2);
+  }
```
Rohit-Kakodkar, line 50: The index ordering is dependent on where this this team lies. We need to have caching for default views on CPUs and coalescing on GPUs. 

Since `LayoutRight` is default for CPUs, to achieve caching on a 2D view I think we'd need to assign every column to individual thread i.e. i1 should increase first. 

For GPUs, LayoutLeft is default, to achieve coalescing on 2D view we'd need to assign every row to individual threads. So i2 needs should increase first. 
```suggestion
  KOKKOS_INLINE_FUNCTION const index_type operator()(const int &i) const {
   if constexpr (std::is_same_v<TeamMemberType::execution_space, Kokkos::DefaultHostExecutionSpace>) {
     const int i1 = i % N1;
     const int i2 = i / N1;
     return index_type(i1, i2);
   } else {
     const int i1 = i / N2;
     const int i2 = i % N2;
     return index_type(i1, i2);
   }
  }
```

include/execution/team_thread_md_range_iterator.hpp
```diff
@@ -0,0 +1,92 @@
...
+      int, Kokkos::extents<std::size_t, rank>,
+      Kokkos::layout_left>; ///< Underlying index type. This
+                            ///< index will be passed to the
+                            ///< closure when calling @ref
+                            ///< kokkos::parallel_for with
+                            ///< this iterator.
+
+  using execution_space =
+      typename base_type::execution_space; ///< Execution space type.
+  KOKKOS_INLINE_FUNCTION const index_type operator()(const int &i) const {
```
Rohit-Kakodkar, line 85: See comment above.

Rohit-Kakodkar (changes requested):

include/execution/thread_md_range_iterator.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <Kokkos_Core.hpp>
+#include <type_traits>
+
+namespace specfem {
+namespace execution {
+
+template <typename TeamMemberType, int... Ranks> class ThreadMDRangeIterator;
+
+template <typename TeamMemberType, int N1, int N2>
+class ThreadMDRangeIterator<TeamMemberType, N1, N2>
```
Rohit-Kakodkar: Should be `TeamThreadMDRangeIterator`

lsawade (approved):

include/execution/chunked_domain_iterator.hpp
icui: Here should be `ngllx` since `iz` range from 0 to `ngllz`

include/execution/team_thread_md_range_iterator.hpp
icui: is `if constexpr (std::is_same_v<TeamMemberType::execution_space, Kokkos::DefaultHostExecutionSpace>)` preferred over `#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)` used in `chunked_domain_iterator.hpp`?

icui: Also, currently Views are not always using the default layout. For example, mesh quadrature View has hard-coded layout right.

Rohit-Kakodkar (approved):

PR #1259, opened 2025-10-02
Conforming-nonconforming kernel agreement test

int-ptr-ptr:
## Description

Please describe the changes/features in this pull request.
~~- [ ] uncomments + populates NCI kernel from #1202~~
~~- [ ] closes dependencies for kernel implementation (#1256, #1257)~~
- [x] adds a test to verify, for conforming mesh(es), that the nonconforming and conforming kernel gives the same result when calling `kernel.update_wavefields<...>()`.

Edit: This will only be the unit test.

## Issue Number

If there is an issue created for these changes, link it here
closes #1258

## Checklist

Please make sure to check developer documentation on specfem docs.

- [ ] I ran the code through pre-commit to check style
- [ ] **THE DOCUMENTATION BUILDS WITHOUT WARNINGS/ERRORS**
- [ ] I have added labels to the PR (see right hand side of the PR page)
- [ ] My code passes all the integration tests
- [ ] I have added sufficient unittests to test my changes
- [ ] I have added/updated documentation for the changes I am proposing
- [ ] I have updated CMakeLists to ensure my code builds
- [ ] My code builds across all platforms

Issue #1256, opened 2025-10-02
Add data_access functions (load_on_device etc.)


int-ptr-ptr:
## Task

- [ ] Create a nonconforming specialization of `specfem::point::coupled_interface` (and/or `specfem::chunk_edge::coupled_interface`), along with corresponding `load_on_device` functions.

## Acceptance Criteria

- [ ] #1265 does not raw-access Kokkos views.

Issue #1226, opened 2025-09-18
Improve compute_intersection / compute_transfer testing


int-ptr-ptr:
Current `nonconforming` framework is minimal, with nonstandard verbose output. `ExpectedGot` could be improved to display arrays / vectors in a layout-preserving fashion, which is helpful for parsing errors.

- [ ] standardize a verbose `specfem::testing::array{n}d` or an `ExpectedGot`-like capture for `specfem::test_fixture` to take the comparisons from the device kernel to host space for necessary printing.
- [ ] Convert typed tests to the parameterized tests, like the new `algorithms/transfer`. If we can create some sort of pattern to capture `test_fixture` modules, that would be great.
- [ ] (OPTIONAL) With the migration of typed tests to regular parameterized tests, we can join together certain test cases fairly easily into one `gtest` test (concat or cartesian product). Do this.
- [ ] Multiple tests use a copy-paste `ChunkEdgeIndex` patch that removes the iterator wrapping for the outer par-for loop. Combine them into one struct.

## Acceptance Criteria

- [ ] tests have nicer naming
- [ ] verbose fail messages provide additional context (transfer function values, etc) in a tabulated way.

PR #1340, opened 2025-11-01
Conforming-nonconforming kernel agreement test

### LINKED ISSUE #1258: Add test for NCI compute_coupling
## Task description

Write a list of things that need to be completed to achieve the goal of this issue.

~~- [ ] For the sake of TDD, write a test for the NCI kernel.~~
The prior test is deprecated. The dG is merged, so we can use passing as the acceptance criterion.

- [ ] Create a test for compute coupling. This should also provide an easy means to create / reference geometry and fields for other interface tests

## Acceptance criteria
~~- [ ] The test provides an expected result to compare to in order to verify NCIs. Test should fail, as kernel is not implemented.~~
- [x] The test passes

int-ptr-ptr:
## Description

(formerly #1259, this PR removes some history, since the "changed files" tab on GitHub was messed up)

~~- [ ] uncomments + populates NCI kernel from #1202~~
~~- [ ] closes dependencies for kernel implementation (#1256, #1257)~~
- [x] adds a test to verify, for conforming mesh(es), that the nonconforming and conforming kernel gives the same result when calling `kernel.update_wavefields<...>()`.

Edit: This will only be the unit test.

## Issue Number

closes #1258

## Checklist

### DIFF
--- a/tests/unit-tests/CMakeLists.txt
+++ b/tests/unit-tests/CMakeLists.txt
@@ -316,16 +316,20 @@ add_executable(
   nonconforming_tests
   nonconforming/reparameterizations/compute_intersection_test.cpp
   nonconforming/reparameterizations/set_transfer_functions_test.cpp
+  nonconforming/kernel_compare/fixture.cpp
+  nonconforming/kernel_compare/kernel_compare_test.cpp
   nonconforming/runner.cpp
 )
 
 target_link_libraries(
   nonconforming_tests
+  io
   mesh
   assembly
   quadrature
   mpi_environment
   kokkos_environment
+  kokkos_kernels
   yaml-cpp
   utilities
   boost
@@ -932,6 +936,7 @@ if(NOT MPI_PARALLEL)
     io
     mesh
     policies
+    nonconforming
   )
 
   foreach(dir_name IN LISTS LINK_DIRS)
--- a/tests/unit-tests/nonconforming/kernel_compare/README.md
+++ b/tests/unit-tests/nonconforming/kernel_compare/README.md
@@ -1,0 +1,33 @@
+# Unit Test: Nonconforming Kernel Comparison
+
+`nonconforming/kernel_compare`
+
+This test verifies the standard acoustic-elastic coupling to the nonconforming implementation on the same mesh. The coupling is done by enforcing traction and displacement continuity in the normal direction on the interface in the weak problem ([Komatitsch, Barnes, Tromp (2000)](https://doi.org/10.1190/1.1444758)).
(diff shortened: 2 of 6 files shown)


int-ptr-ptr:
I do feel like this test could be difficult to understand, and might benefit from a rework (say issue #1226). Otherwise, it may be that #1226 is not necessary.

lsawade (commented):
I think the idea of this test is great. I'm not quite understanding though of the core part where you update the field. You 
- clear the fields (zero them out)
- compute a single time step (update_wavefield),  which 
  - couples zero's (??)
  - adds some source term to the acceleration
  - computes some stiffness etc.

Meaning that in the end unless the source element is on the boundary you are not testing that a field is actually traversing the interface, but only that the source term is added in the same fashion in both kernels. Am I misunderstanding this?

int-ptr-ptr:
> I think the idea of this test is great. I'm not quite understanding though of the core part where you update the field. You
> 
>     * clear the fields (zero them out)
> 
>     * compute a single time step (update_wavefield),  which
>       
>       * couples zero's (??)
>       * adds some source term to the acceleration
>       * computes some stiffness etc.
> 
> 
> Meaning that in the end unless the source element is on the boundary you are not testing that a field is actually traversing the interface, but only that the source term is added in the same fashion in both kernels. Am I misunderstanding this?

There are no source terms added directly to the acceleration. The fields are cleared *except* for a single shape function. See the [`README.md`](https://github.com/int-ptr-ptr/SPECFEMPP/tree/dg2d-kerneltest/tests/unit-tests/nonconforming/kernel_compare) for the test (already in the PR).

In essence, [these](https://github.com/int-ptr-ptr/SPECFEMPP/blob/dg2d-kerneltest/tests/unit-tests/nonconforming/kernel_compare/kernel_compare_test.cpp#L371-L387) are the lines that correspond to the displacement field setting. (also below for your convenience)

```c++
            for (const auto &[nc_iglob, inds] : nodes_to_check) {
              const auto &[nc_index, c_index] = inds;
              for (int icomp = 0; icomp < PointDisplacementType::components;
                   icomp++) {
                displacement(icomp) = 1;                            // <<< HERE
                // ===================================
                // (foreach degree of freedom i) then:
                // ===================================


                clear_field<specfem::point::acceleration>(nc_assembly);
                clear_field<specfem::point::acceleration>(c_assembly);


                specfem::assembly::store_on_host(nc_index, nc_field,
                                                 displacement);
                specfem::assembly::store_on_host(c_index, c_field,
                                                 displacement);
```

PR #1352, opened 2025-11-05
Add `KOKKOS_PATH` flag to CMake

Rohit-Kakodkar:
## Description

This PR adds support for using a local installation of Kokkos when building SPECFEM++

## Issue Number

## Checklist

### DIFF
--- a/cmake/kokkos.cmake
+++ b/cmake/kokkos.cmake
@@ -1,5 +1,11 @@
 
 
+if (DEFINED Kokkos_ROOT)
+    message(STATUS "Using user-defined Kokkos root: ${Kokkos_ROOT}")
+    find_package(Kokkos REQUIRED PATHS ${Kokkos_ROOT} NO_DEFAULT_PATH)
+    return()
+endif()
+
 message(STATUS "Configuring Kokkos library...")
 
 # Prepend the CMAKE_MESSAGE_INDENT variable to ensure proper indentation in messages
--- a/cmake/kokkos.cmake
+++ b/cmake/kokkos.cmake
@@ -27,26 +27,33 @@ if (CMAKE_VERSION VERSION_GREATER "3.30.0")
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


icui (approved):

cmake/kokkos.cmake
```diff
@@ -27,26 +27,33 @@ if (CMAKE_VERSION VERSION_GREATER "3.30.0")
     endif()
 endif()
 
-set(KOKKOS_VERSION "4.7.00")
+if (DEFINED KOKKOS_PATH)
```
lsawade, line 36: This is non-standard. the variable should be `Kokkos_ROOT` and then you should be able to use: 
```cmake
find_package(Kokkos REQUIRED)

add_executable(main main.cpp)
target_link_libraries(main Kokkos::kokkos)
```
this will then also define `Kokkos_FOUND` etc. which can then be used.

lsawade: I also tested this, so this works

lsawade (changes requested):

cmake/kokkos.cmake
Rohit-Kakodkar: There are fundamentally 2 ways for building kokkos, (see at the bottom of this [slides](https://github.com/kokkos/kokkos-tutorials/blob/main/LectureSeries/KokkosTutorial_01_Introduction.pdf))
- as a pre installed library (by providing a kokkos root path)
- as part of your own library (by adding kokkos as a subdirectory)

Using a pre-installed library would mean that we would use any default parameters that the cluster administrators installed the library with (which would affect performance)

lsawade: I misunderstood what we discussed for this issue.

You want to give an external source code path and still build it as part of the project.

I think we should still provide the find_kokkos interface and `SPECFEM_KOKKOS_FORCE_INSTALL`, but maybe a different default behavior/option val for Kokkos.

PR #1359, opened 2025-11-06
1178 final - finalize displacement tests

### LINKED ISSUE #1178: Make adjancency graph as default for computing different mesh properties
## Task description

We currently store mesh connectivity within an adjacency graph. This is fairly recent feature, we need to update several methods to use this feature 

- Remove `write_adjacency_map` flag. (Effort Level : 1)
   - Make computation and writing of the adjacency map the default for the mesher 
- Update all databases within displacement tests to store the adjacency map generated by the mesher. The new mesher generates an adjacency map by default. (Effort Level : 3)
   - Write snakemake within all displacement tests that generates the traces from provenance
- Update assembly computation to only use adjacency maps (Effort Level : 1)
   - Remove  `assemble_legacy` function used to compute assembly
- Update coupled interface storage (Effort Level: 1)
   - Store coupled interfaces as `weakly_coupled` connection within 2 elements 

Note to search for where things need to change `grep ADJ_GRPAH_DEFAULT`

## Acceptance criteria
- [ ] Displacement tests pass

icui:
## Description

- Add a new `HomogeneousAcousticDomainCompositeStaceyDirichletBC` since the old provenance is missing
- Add workflow for `database.bin` for Cosserat domain
- For now I'm not adding finite difference trace to the workflow because it is CUDA only
- Re-generate all `database.bin` with updated format (deleting the `write_adjacency_map` line)

## Issue Number

Closes #1178 

## Checklist


src/io/mesh/impl/fortran/dim2/read_adjacency_graph.cpp
```diff
@@ -14,21 +14,6 @@ specfem::io::mesh::impl::fortran::dim2::read_adjacency_graph(
...
-    stream.seekg(current_position);
-    read_adjacency_graph = false;
-  }
-  if (!read_adjacency_graph) {
-    specfem::mesh::adjacency_graph<specfem::dimension::type::dim2>
-        empty_graph{};
-    return empty_graph;
-  }
-
   specfem::mesh::adjacency_graph<specfem::dimension::type::dim2> graph(nspec);
```
int-ptr-ptr, line 17: I forsee issues where someone would want to use a specfem2d database and miss any documentation saying you can't do that. It may not be necessary at the moment, but would it be beneficial to have a more verbose fail message, here?

If this was not the end of specfem2d database, then such a check would require a fiducial bitstring or something of the sort, which may not be worth it. But, since a `specfem2d` database should give an EOF here, while a `specfempp` database would keep going, I think it would be better to keep the `try-catch` on `fortran_read_line`, but throw another `std::runtime_error("unexpected EOF: make sure the database is generated with the SPECFEM++ meshfem2D code")`.

int-ptr-ptr (commented):
I think this makes sense, but I might need to see it in action.

Thank you!

Rohit-Kakodkar (commented):
> I think this makes sense, but I might need to see it in action.
> 
> Thank you! 

The fortran read line function should generate end of line error when trying to read line here.

lsawade (approved):

Rohit-Kakodkar (approved):

int-ptr-ptr (approved):

int-ptr-ptr:
> The fortran read line function should generate end of line error when trying to read line here.

Yes, but an elaboration on that error message would be helpful to diagnose issues. I've noticed a lot of the error messages broadly for the database and `config.yaml` reading are not very verbose.

It's not a high priority, but I think it's worth making a note of. You've probably also encountered that when updating the `yaml` files after a parser change.

PR #1424, opened 2025-11-21
Issue 1401 - Refactor non conforming interfaces accessors

Rohit-Kakodkar:
## Description

- [x] Renamed coupled interfaces to conforming interfaces (Future PR will decouple conforming and nonconforming containers & kernel calls)
- [x] Refactored chunk edge and point non-conforming interfaces accessor. The new implementation using accessor packing infrastructure so multiple accessors can be packed and passed across the code base. This should make implementation of new flux schemes much easier
- [x] Refactors non-conforming interface container access function so that they can interact with packed accessors
- [x] Makes changes to rest of codebase required by above changes

## Issue Number

## Checklist


include/medium/compute_coupling.hpp
```diff
@@ -82,6 +82,28 @@ compute_coupling(const CoupledInterfaceType &interface_data,
...
+    DimensionDispatch dimension, ConnectionDispatch connection,
+    InterfaceDispatch interface,
+    const std::integer_sequence<std::size_t, Is...> /*unused*/,
+    const IndexType &index, const InterfaceDataType &interface_data,
+    const CoupledFieldType &coupled_field, SelfFieldType &self_field) {
+
+  compute_coupling(dimension, connection, interface, index,
+                   static_cast<const std::tuple_element_t<
+                       Is, typename InterfaceDataType::packed_accessors> &>(
+                       interface_data)...,
```
int-ptr-ptr, line 101: I like the idea of pack-folding here so that the per-interface / per-flux-scheme code gets more transparent with what is needed, but I worry that if the ordering is not made explicit somewhere, we could run into an issue of mixing up accessors. Placing `check_compatibility` `static_assert`s would help, but seems a bit ugly IMO, since the expansion happens deep in the `impl` namespace.

core/specfem/chunk_edge/nonconforming_interface.hpp
```diff
@@ -0,0 +1,287 @@
...
+      public Accessors... {
+
+  constexpr static auto dimension_tag =
+      std::tuple_element_t<0, std::tuple<Accessors...> >::dimension_tag;
+  constexpr static auto interface_tag =
+      std::tuple_element_t<0, std::tuple<Accessors...> >::interface_tag;
+  constexpr static auto boundary_tag =
+      std::tuple_element_t<0, std::tuple<Accessors...> >::boundary_tag;
+  constexpr static size_t n_accessors = sizeof...(Accessors);
+  using packed_accessors = std::tuple<Accessors...>;
```
int-ptr-ptr, line 217: While fine for now, I do worry about fixed orderings and explicit getting, though these are more focused on the test-writing process.

For the former, we need to enforce that the `coupling_terms_pack` and `integral_data_pack` orderings match those of the medium `coupling_terms` impl functions. It's okay as WET code, but feels really ugly to me. Not to mention, you are required to use the specific pack. With the former code, we could use the universal `nonconforming_coupled_interface` accessor for any medium and flux scheme, which would be nicer than recreating a new pack for each scheme.

Regarding the latter: Before, it was simple to test the `transfer_function_self`, for example, since you would just use that field name, but now, you would have to iterate the pack and return the first accessor with `DataClassType::transfer_function_self` (granted, this could be a templated method).

int-ptr-ptr (approved):

core/specfem/chunk_edge/nonconforming_interface.hpp
Rohit-Kakodkar: I believe the way to verify that is within the `medium::coupling_terms` implementation functions for different flux schemes. For example, in this function [here](https://github.com/PrincetonUniversity/SPECFEMPP/blob/0ea6a0a24bd4ccf9d895ed6c066cf6ee7e6b90db/include/medium/dim2/coupling_terms/acoustic_elastic.hpp#L63), I could have added a static assert. 

```cpp
   static_assert(specfem::data_access::is_transfer_function_self<TransferFunctionSelf>::value, ...)
```

The lowest-level implementation functions are always unpacked. Granted, you wouldn't be able to verify this in load or store functions, but you shouldn't need to.

core/specfem/assembly/nonconforming_interfaces/dim2/impl/nonconforming_interface.hpp
```diff
@@ -103,177 +103,201 @@ struct interface_container<specfem::dimension::type::dim2, InterfaceTag,
...
+            typename AccessorType, std::size_t... Is>
+  KOKKOS_FORCEINLINE_FUNCTION void
+  impl_load_expand(const DispatchType dispatch, const IndexType &index,
+                   AccessorType &accessor,
+                   std::index_sequence<Is...> /* index sequence */) const {
+    impl_load_after_expansion<on_device>(
+        dispatch, index,
+        static_cast<typename std::tuple_element_t<
+            Is, typename AccessorType::packed_accessors> &>(accessor)...);
+  }
```
lsawade, line 275: cool.

lsawade (approved):

icui (approved):

PR #1431, opened 2025-11-25
Fix Cuda compilation 3D and Updates chunksizes

lsawade:
## Description

Once that was fixed, I was able to update the default chunk config for not only 3D but also 2D. So the computation should be slightly faster now in 

### Discussion point

Seismograms computation for 2 and 3D have drastically different chunksizes, which begs the question whether we should actually create separate values for dimension and kernel computation on CUDA. That is,

- chunksize_mass_{2d,3d}
- chunksize_seis_{2d,3d}
- chunksize_stif_{2d,3d}

etc.

## Commits

- There was a bug that required the allocation of memory during cuda execution. Once that was fixed, I was able to update the default chunk config. (96b15ddc)

## Issue Number

## Checklist

### DIFF
--- a/src/enumerations/dim3/mesh_entities.cpp
+++ b/src/enumerations/dim3/mesh_entities.cpp
@@ -204,7 +204,7 @@ specfem::mesh_entity::element<specfem::dimension::type::dim3>::element(
   // ymin = front, ymax = back
 
   face_coordinates[specfem::mesh_entity::dim3::type::bottom] =
-      [*this](const int ipoint, const int jpoint) {
+      [this](const int ipoint, const int jpoint) {
         const int iz = 0;
         const int iy = ipoint;
         const int ix = jpoint;
@@ -212,7 +212,7 @@ specfem::mesh_entity::element<specfem::dimension::type::dim3>::element(
       };
 
   face_coordinates[specfem::mesh_entity::dim3::type::top] =
-      [*this](const int ipoint, const int jpoint) {
+      [this](const int ipoint, const int jpoint) {
         const int iz = this->ngllz - 1;
         const int iy = ipoint;
         const int ix = jpoint;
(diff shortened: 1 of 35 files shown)


src/enumerations/dim3/mesh_entities.cpp
```diff
@@ -204,143 +204,143 @@ specfem::mesh_entity::element<specfem::dimension::type::dim3>::element(
   // ymin = front, ymax = back
 
   face_coordinates[specfem::mesh_entity::dim3::type::bottom] =
-      [*this](const int ipoint, const int jpoint) {
+      [this](const int ipoint, const int jpoint) {
```
Rohit-Kakodkar: I know we spoke about this, you should capture ngllz/nglly/nglly .

Rohit-Kakodkar (approved):

src/enumerations/dim3/mesh_entities.cpp
lsawade: How about a member function? I submitted a revision. CPU test is now also at 1s. does it need to be a map? If yes, why?

lsawade:
@icui @Rohit-Kakodkar , as Rohit immediately thought correctly, this indeed reduced CPU run time as well to just above a second for the test since it's tiny. CPU/GPU speedup is 12x (solver time only, ran 2000 time steps), which is literally what Rohit said earlier today.

src/enumerations/dim3/mesh_entities.cpp
Rohit-Kakodkar: It doesn't need to be a map; it can be a member function that takes a `mesh_entity` argument—then the function can use switch case to implement the logic.

icui (approved):

Rohit-Kakodkar (approved):

PR #1444, opened 2025-12-04
Change signed wavefields to B-R colors

int-ptr-ptr:
## Description

For wavefields
```c++
        wavefield_type == specfem::wavefield::type::pressure ||
        wavefield_type == specfem::wavefield::type::rotation ||
        wavefield_type == specfem::wavefield::type::intrinsic_rotation ||
        wavefield_type == specfem::wavefield::type::curl
```
a blue-red display scheme is used for `plot_wavefield`.

## Issue Number

## Checklist

### DIFF
--- a/core/specfem/periodic_tasks/plot_wavefield.cpp
+++ b/core/specfem/periodic_tasks/plot_wavefield.cpp
@@ -901,15 +901,17 @@ void specfem::periodic_tasks::plot_wavefield::initialize_display(
   } else {
     double range[2];
     scalars->GetRange(range);
-    double abs_max = std::max(std::abs(range[0]), std::abs(range[1]));
+    double abs_max = (std::abs(range[0]) + std::abs(range[1])) / 2.0;
     this->wavefield_mapper->SetScalarRange(-abs_max, abs_max);
     this->lut->SetRange(-abs_max, abs_max);
 
     for (int i = 0; i < 256; ++i) {
       double t = static_cast<double>(i) / 255.0;
-      double transparency = 2.0 * std::abs(0.5 - t);
-      double step = t > 0.5 ? 1.0 : 0.0;
-      this->lut->SetTableValue(i, step, 0.0, 1.0 - step, transparency);
+      double transparency = this->sigmoid(2.0 * std::abs(0.5 - t));
+      double blue_frac = std::min(2.0 - 2.0 * t, 1.0);
+      double red_frac = std::min(2.0 * t, 1.0);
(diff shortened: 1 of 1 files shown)


Rohit-Kakodkar (approved):

int-ptr-ptr:
One thing that I am not too happy about is how the scaling changes each frame, so a wave that hits a Neumann boundary doubles the table range. Since this solution does not use a sigmoid, it dims the colors everwhere.

That being said, I don't know of any good way to handle that, since the previous sigmoid is way too steep, and acts more like a binary filter, but I think that should be handled in a later PR.

lsawade:
Does the colorscheme look like this? I'm imagining a double-sided sigmoid where transparency is 1.0, when the wavefield is close to zero and 0.0 towards B-R ends of the colorscheme (opposite of alpha value). It does not look like it to me from the code.

image

If not, how about something like this:

```python
def sigmoid(x, b):
     return 1/(1 + np.exp(-x-b))

def doublesigmoid(x, a, b):
     return (sigmoid(x, a) + (1 - sigmoid(x,b)))/2
plt.close('all')
x= np.linspace(-1, 1, 1000)
y = doublesigmoid(x*10,-0.1, 0.1)
y= y-y.min(); y = y/y.max()
plt.plot(x, y)
```

Screenshot 2025-12-09 at 7 40 18 AM

Or, for more zeros around zero:
```python
plt.close('all')
x= np.linspace(-1, 1, 1000)
y = doublesigmoid(x*20,-10, 10)
y= y-y.min()
y = y/y.max()
plt.plot(x, y)
```
Screenshot 2025-12-09 at 7 45 06 AM

core/specfem/periodic_tasks/plot_wavefield.cpp
```diff
@@ -881,16 +886,31 @@ void specfem::periodic_tasks::plot_wavefield::initialize_display(
...
+    scalars->GetRange(range);
+    double abs_max = std::max(std::abs(range[0]), std::abs(range[1]));
+    this->wavefield_mapper->SetScalarRange(-abs_max, abs_max);
+    this->lut->SetRange(-abs_max, abs_max);
+
+    for (int i = 0; i < 256; ++i) {
+      double t = static_cast<double>(i) / 255.0;
+      double transparency = 2.0 * std::abs(0.5 - t);
+      double step = t > 0.5 ? 1.0 : 0.0;
+      this->lut->SetTableValue(i, step, 0.0, 1.0 - step, transparency);
```
lsawade: see comment

lsawade (changes requested):

icui (approved):

PR #1481, opened 2025-12-16
migrated fixtures, added power tests

int-ptr-ptr:
## Description

- `TransferFunction2D` and `EdgeFunction2D` moved to `utilities` in namespace `specfem::test::fixture`
- Tools to create `TransferFunctionInitializer2D` from quadrature point lists
- Added 2 more `TransferFunctionInitializer2D` structs
- Added `EdgeFunctionInitializer2D` struct: `FromEdgeField`, which populates `EdgeFunction2D` from quadrature points in `TransferFunction2D`
- re-introduced power tests $f(\xi) = \xi^k$, where $k$ is small enough for the quadrature to be exact.

## Issue Number

## Checklist


Rohit-Kakodkar:
These test cases are great, especially because they check against an analytical solution.

However, I think the PR makes the initial implementation less transparent. After these changes, it’s less clear how one would add another test case because:
- The test cases do not follow a consistent pattern in their initialization.
- The test cases are not self-contained.
- `initializer_value_check` is opaque.

I would suggest the following update
```cpp
namespace AnalyticalSolutions {
class PowerTests {
}; // rename this if you need
}

namespace TransferFunctionInitializer {
class Zero {};
class Gll1_to_GLL2 : public AnalyticalSolutions::PowerTests {
   ...
};
....
}

template <typename Initializer, std::enable_if_t<std::is_base_of<AnalyticalSolutions::PowerTest> 
auto init_transfer_function
{
....
}

// Make similar changes. 
```

int-ptr-ptr:
@Rohit-Kakodkar  I've taken these suggestions into mind. With inheritance, I hope it becomes more clear. Each type ensures that its initializer inherits from the corresponding `...Initializer...D::Base` rather than verifying through `initializer_value_check`. For example, when creating a new `EdgeField2D` case, one can take after `Uniform` and explicitly make it, or if there is an analytical formula, `FromAnalyticalField` could be used, which is templated off of the analytical formula and quadrature rule (both do the "type checking" in the `using` commands). Rather than doing some minor guesswork at omitted parameters, I'll just let the compiler scream at the user if they leave out a field.

```c++
struct Uniform : Base {
  static constexpr int num_edges = 1;
  static constexpr int num_components = 1;
  static constexpr int nquad_edge = 5;
  static void init_function(EdgeField2D<Uniform> &edge_function) {
    for (size_t i = 0; i < num_edges; ++i) {
      for (size_t j = 0; j < nquad_edge; ++j) {
        for (size_t k = 0; k < num_components; ++k) {
          edge_function(i, j, k) = 1;
        }
      }
    }
  }

  static std::string description() {
    return std::string("ngll = 5 field of all ones");
  }
};

template <typename AnalyticalFieldInitializer, typename EdgePointsInitializer>
struct FromAnalyticalField : Base {
  using AnalyticalField = AnalyticalField1D<AnalyticalFieldInitializer>;
  using EdgeQuadrature = QuadratureRule<EdgePointsInitializer>;

  static constexpr int num_edges = AnalyticalField::num_edges;
  static constexpr int num_components = AnalyticalField::num_components;
  static constexpr int nquad_edge = EdgeQuadrature::nquad;
  static constexpr std::array<type_real, nquad_edge> edge_quadrature_points =
      EdgeQuadrature::quadrature_points;

  static std::string description() {
    return std::string("EdgeField-initialized function (description: \"") +
           AnalyticalField::description() + "\")";
  }

  static void init_function(EdgeField2D<FromAnalyticalField> &edge_function) {
    for (size_t i = 0; i < num_edges; ++i) {
      for (size_t j = 0; j < nquad_edge; ++j) {
        for (size_t k = 0; k < num_components; ++k) {
          edge_function(i, j, k) =
              AnalyticalField::evaluate(i, edge_quadrature_points[j], k);
        }
      }
    }
  }
};
```

The idea was to allow self-contained cases, but in the event that there is a useful construct (say, `GLL5`) that could be used in multiple tests in a consistent way (e.g. sampling at quadrature points), it would be a separate "fixture" that can be brought in.

I feel like this is a fairly straightforward pattern, but please let me know if it's not.

tests/unit-tests/algorithms/dim2/transfer.cpp
```diff
@@ -63,275 +69,160 @@ constexpr static auto interface_tag =
...
+  TransferFunctionTest2D()
+      : transfer_function(TransferFunctionInitializer()),
+        edge_field(FunctionInitializer()) {}
+
+  TransferFunction2D transfer_function; /**< Test transfer function */
+  EdgeField2D edge_field;               /**< Test field */
+
+  std::vector<std::array<std::array<type_real, EdgeField2D::num_components>,
+                         TransferFunction2D::nquad_intersection> >
+  expected_solution() const {
```
Rohit-Kakodkar: I am still not understanding why this can't be 

```cpp
template <typename TransferFunction2D, typename EdgeFunction2D, std::enable_if_t<std::base_of_v<TransferFunction2D, PowerTest> && std::enable_if_t<std::base_of_v<TransferFunction2D, PowerTest>>
auto compute_transfer(const TransferFunction2D & /*unused*/, const TransferFunction2D & /*unused*/) {
    const size_t power = EdgeFunction2D::power;
    const auto intersection_points = TransferFunction2D::intersection_points;
    // Do the loop expliciltly.
}

Comment (Rohit-Kakodkar) [reply tests/unit-tests/algorithms/dim2/transfer.cpp]: Making this into a class makes it difficult to add more test cases in the future. For most of our developers, this is the first time they see C++ code and definitely the first time they see a templated C++ code. The easier we make the developer experience, the better.

File: tests/unit-tests/utilities/include/fixture/nonconforming_interface/edge_field.hpp [hunk 2]
```diff
@@ -0,0 +1,137 @@
...
+   * @param initializer Initialization strategy
+   */
+  EdgeField2D(const FunctionInitializer &initializer) {
+    for (int ielem = 0; ielem < num_edges; ++ielem) {
+      for (int ipoint_edge = 0; ipoint_edge < nquad_edge; ++ipoint_edge) {
+        for (int icomp = 0; icomp < num_components; ++icomp) {
+          (*this)(ielem, ipoint_edge, icomp) = 0;
+        }
+      }
+    }
```
Rohit-Kakodkar: If `init_function` is responsible for initializing this field. This block is redundant.

tests/unit-tests/utilities/include/fixture/nonconforming_interface/analytical_field.hpp
Rohit-Kakodkar, line 1: I think these two functions can be removed in favour of having an inline implementation of the functions.

tests/unit-tests/utilities/include/fixture/nonconforming_interface/quadrature_rule.hpp
Rohit-Kakodkar, line 1: These classes can be deprecated in favor of having them inline within the transfer function initializer.

Rohit-Kakodkar (commented):
I don't know how you are using inheritance in the PR yet. Seems like all initializers are derived from `Base`. 

IMO, the implementation can be simplified further with inheritance.

Rohit-Kakodkar:
@int-ptr-ptr seems like this closed when I merged the PRing branch into devel. Can you re-request this PR? Sorry.

lsawade (commented):
@Rohit-Kakodkar are you happy with these updates?

tests/unit-tests/algorithms/dim2/transfer.cpp
```diff
@@ -267,41 +73,103 @@ template <typename Initializer> struct EdgeFunction2D {
...
+        specfem::test::fixture::EdgeFunctionInitializer2D::
+            FromAnalyticalFunction<AnalyticalFunction, EdgeQuadraturePoints> >
+        &field) {
+  using TransferFunction2D = specfem::test::fixture::TransferFunction2D<
+      specfem::test::fixture::TransferFunctionInitializer2D::
+          FromQuadratureRules<EdgeQuadraturePoints,
+                              IntersectionQuadraturePoints> >;
+  using EdgeFunction2D = specfem::test::fixture::EdgeFunction2D<
+      specfem::test::fixture::EdgeFunctionInitializer2D::FromAnalyticalFunction<
+          AnalyticalFunction, EdgeQuadraturePoints> >;
```
Rohit-Kakodkar: his syntax here could be much shorter and more clear with following change. 

```suggestion
template <typename TransferFunction2d, typename EdgeFunction2D>
std::enable_if_t<is_analytical2d_v<EdgeFunction2D>, std::vector<std::array<std::array<type_real, EdgeFunction2D::num_components>,
                       TransferFunction2D::nquad_intersection> >
expected_solution(
    const TransferFunction2d &transfer_function, const EdgeFunction2d &edge_function)
```

Rohit-Kakodkar: Check the syntax. Also add `is_analytical2d_v` helper.

tests/unit-tests/algorithms/dim2/transfer.cpp
```diff
@@ -389,17 +276,42 @@ struct TransferFunctionTest2D : public ::testing::Test {
...
 };
 
+using specfem::test::fixture::AnalyticalFunctionType1D::Power;
+using specfem::test::fixture::EdgeFunctionInitializer2D::FromAnalyticalFunction;
+using specfem::test::fixture::QuadraturePoints::Asymm4Point;
+using specfem::test::fixture::QuadraturePoints::Asymm5Point;
+using specfem::test::fixture::QuadraturePoints::GLL1;
+using specfem::test::fixture::QuadraturePoints::GLL2;
+using specfem::test::fixture::TransferFunctionInitializer2D::
+    FromQuadratureRules;
```
Rohit-Kakodkar: Why not ``specfem::test::fixture``

tests/unit-tests/utilities/include/fixture/nonconforming_interface/analytical_function.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "initializers.hpp"
+#include "specfem_setup.hpp"
+
+namespace specfem::test::fixture {
```
Rohit-Kakodkar: Can we coalesce this namespace ``specfem::test_fixture``

tests/unit-tests/utilities/include/fixture/nonconforming_interface/edge_function.hpp
```diff
@@ -0,0 +1,164 @@
...
+        const auto eval =
+            AnalyticalFunction::evaluate(edge_quadrature_points[j]);
+        // if num_components is 1, we may see a type_real returned, rather than
+        // an array.
+        if constexpr (std::is_arithmetic_v<decltype(eval)>) {
+          edge_function[i][j][0] = eval;
+        } else {
+          for (size_t k = 0; k < num_components; ++k) {
+            edge_function(i, j, k) = eval[k];
+          }
```
Rohit-Kakodkar: Not sure why you need this. If the number of components = 1, then the analytical function should know it and return a ``std::array<type_real,1>``

tests/unit-tests/utilities/include/fixture/nonconforming_interface/edge_function.hpp
```diff
@@ -0,0 +1,164 @@
...
+    return edge_function;
+  }
+
+  static std::string description() {
+    return std::string("ngll = 5 field of all ones");
+  }
+};
+
+template <typename AnalyticalFunctionType, typename EdgePoints>
+struct FromAnalyticalFunction : EdgeFunctionInitializer2D {
```
Rohit-Kakodkar, line 46: This should be a base of ``AnalyticalFunction2D``.

Rohit-Kakodkar (changes requested):
Mostly looks good. Minor changes requested.

tests/unit-tests/algorithms/dim2/transfer.cpp
int-ptr-ptr: I'm not sure what you mean here. These `using` statements just allow for single-name references in the tuples, which I feel is more readable, since it gets collapsed into 2 lines per entry.

tests/unit-tests/utilities/include/fixture/nonconforming_interface/edge_function.hpp
int-ptr-ptr: This is written so that `AnalyticalFunctionND` has `N` parameters. Since we are defining the function off of edge local coordinates, `EdgeFunctionInitializer2D` expects an analytical function defined on codimension 1 (`N=1`).

tests/unit-tests/algorithms/dim2/transfer.cpp
Rohit-Kakodkar: Sorry I meant ``using namespace specfem::test::fixture`` so that you dont have to individually import each type.

tests/unit-tests/utilities/include/fixture/nonconforming_interface/edge_function.hpp
Rohit-Kakodkar: Sure, then I guess it should be a base of `AnalyticalFunction`. It seems to me that the tests you are adding right now - you know the exact solution because the edge function is analytical. Having it as a base would be a way to categorize your tests. 

The base checking will then be used in the expected solution SFINAE.

tests/unit-tests/algorithms/dim2/transfer.cpp
int-ptr-ptr: That is what I had tried to do initially. The issue was that the function call was ambiguous -- I added a negated std::enable_if_t on the base just now, which solved that.

tests/unit-tests/algorithms/dim2/transfer.cpp
```diff
@@ -63,245 +65,113 @@ constexpr static auto interface_tag =
...
-                   memory_space>;
-
-public:
-  /**
-   * @brief Construct transfer function with initializer.
-   * @param initializer Initialization strategy
-   */
-  TransferFunction2D(const Initializer &initializer)
-      : _transfer_function(init_transfer_function<Initializer>()) {}
+template <typename TransferFunction2D, typename EdgeFunction2D>
```
Rohit-Kakodkar, line 86:
```suggestion
// TODO(Rohit : CPP20 update) Update this SFINAE with concepts
template <typename TransferFunction2D, typename EdgeFunction2D>
```

Rohit-Kakodkar (approved):

tests/unit-tests/algorithms/dim2/transfer.cpp
Rohit-Kakodkar: Commiting this comment for future change.

icui (approved):

lsawade (approved):
This looks great! Thanks for making all the updates!

PR #1665, opened 2026-02-23
[Draft] Attenuation implementation for discussion.

lsawade:
## Description

This is a very rough draft of what is needed for attenuation and which computations need to take place.

I made this as a point of reference for discussion, so that we can think about how to, in particular, implement the strain time taylor expansion.

### DIFF
--- a/core/specfem/compute/impl/compute_stiffness_interaction.tpp
+++ b/core/specfem/compute/impl/compute_stiffness_interaction.tpp
@@ -208,8 +208,12 @@ int specfem::compute::impl::compute_stiffness_interaction(
                 specfem::assembly::load_on_device(index, attenuation,
                                                     point_integration_factors);
 
-                // Compute the time derivative of the field derivatives using the velocity field
-                PointFieldStrain point_strain(field_derivatives);
+                // Load previous strain for memory variable update
+                PointFieldStrain point_strain;
+                specfem::assembly::load_on_device(index, attenuation, point_strain);
+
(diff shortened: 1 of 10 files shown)


Rohit-Kakodkar (commented):
This looks great. I am happy to approve this as is if it weren't for performance degradation we's see due to need to computing ``dv``.

One way of solving this would be to introduce "GradientFieldPack" akin to "NonconformingAccessorPack" which specifies which fields we have to take gradients of. Then we can statically unpack the struct which shouldnt affect performance. 

The other thing that could be streamlined is - we merge ``compute_stress_relaxation``, ``update_memory_variable`` within a single function. This would make it more readable.

lsawade:
Yes, I was thinking gradient field pack as well. I did not think about fusing the attenuation functions together since they are very spread out in the fortran code as well. But one function that adds to stress and integrates R should be doable!

core/specfem/compute/impl/compute_stiffness_interaction.tpp
```diff
@@ -135,17 +141,21 @@ int specfem::compute::impl::compute_stiffness_interaction(
           const auto &chunk_index = chunk_iterator_index.get_index();
           const auto team = chunk_index.get_policy_index();
           ChunkElementFieldType element_field(team.team_scratch(0));
+          ChunkElementFieldDotType element_field_dot(team.team_scratch(1));
```
Rohit-Kakodkar, line 144:
```suggestion
          ChunkElementFieldDotType element_field_dot(team.team_scratch(0));
```

PR #1669, opened 2026-02-25
Update the gradient algorithm to return the jacobian if the callable …

lsawade:
## Description

Another couple of percent of performance. This should affect GPU performance as well. Since it removes 4 type real loads

`specfem::algorithms::gradient` now optionally forwards the point Jacobian matrix to the callback                                                                                                  
   
The gradient algorithm previously always loaded a jacobian_matrix (without the Jacobian scalar) internally and passed only the gradient tensor df to the callback. Callers that also needed the Jacobian (e.g., to compute stress) had to perform a second redundant load_on_device call.

This commit uses if constexpr + std::is_invocable_v to detect at compile time whether the callback accepts a third argument of type point::jacobian_matrix<..., store_jacobian=true, ...>. If so, the gradient function loads the richer Jacobian type (with the scalar Jacobian) once and forwards it directly to the callback, eliminating the duplicate load.

The compute_stiffness_interaction caller is updated to use the new three-argument signature, removing its now-redundant load_on_device call. The two-argument signature remains fully supported, so existing callers are unaffected.

Applied to both the 2D and 3D overloads of gradient.

## Issue Number

-/-

## Checklist

### DIFF
--- a/core/specfem/algorithms/gradient.hpp
+++ b/core/specfem/algorithms/gradient.hpp
@@ -203,14 +203,26 @@ gradient(const ChunkIndexType &chunk_index,
 
   using datatype = typename VectorFieldType::simd::datatype;
 
-  static_assert(
+  using JacobianWithJacobianType =
+      specfem::point::jacobian_matrix<specfem::element::dimension_tag::dim2,
+                                      true, using_simd>;
+
+  // If the callback accepts a third argument (the point jacobian matrix with
+  // jacobian scalar), load with store_jacobian=true and forward it — avoids a
+  // redundant load_on_device call inside the callback.
+  constexpr bool pass_jacobian =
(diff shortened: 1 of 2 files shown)


lsawade:
FYI I'm not really a fan of this, see copilots comments. But maybe we can think about how to improve the jacobian load here

Rohit-Kakodkar (commented):
Does this actually affect performance? From my intuition, since jacobian matrix was loaded right before the callback functor, it should lie at least in the L1 cache. So when the functor calls it again it should be available with very low latency.

lsawade:
I got a few % on mac but maybe that's just variance in CPU performance. I have not run the actual benchmarks!

lsawade:
Cannot reproduce.

PR #1670, opened 2026-02-25
Updated the inner loop of the divergence for fewer accesses.

lsawade:
## Description

Updated the inner divergence loop for another couple of percent of performance on Macos

Performance on CPU runtime should be almost 50% smaller. May have impact on GPU. 
But unclear.

## Issue Number

-/-

## Checklist


Rohit-Kakodkar (approved):
This LGTM. However, not intutive on why we get performance gains. 

The compiler should have been able to see this optimization. Both ``ngll`` and ``ncomp`` are constexpr vars, the loops should have been unrolled and values precalculated.

lsawade:
cannot reproduce. Likely local CPU variance. I ran these PRs quite a bit before submitting and ran them against previous branch compiled separately, but I guess CPU variance is about 5% on macos

Issue #1707, opened 2026-03-04
Update specfem_config/runtime configuration


icui:
- mesher writes to folder if nproc>1 in binary
- mesher writes to file if nproc=1 in binary
- Broadcast parameters to all procs

PR #1722, opened 2026-03-11
1707 - Update IO for 2D / 3D  mesh with MPI

icui:
## Description

- Use name {database_filename}/proc_{n}.bin for MPI writing
  - for `foo/bar.bin`, meshfem will from to `foo/bar/proc_1.bin`, `foo/bar/proc_2.bin`, ... if `NPROC > 1`
- Update fortran mesh writer for MPI file name
- Update mesh reader for MPI file name 

## Issue Number

Adds to #1707

## Checklist


lsawade (approved):

Rohit-Kakodkar (commented):
Is proc{n}_{database_filename} written to the same folder as database_filename? It could pollute the directory with unnacessary files. Maybe database_filename/proc{n}.bin would be a better solution?

icui:
> Is proc{n}_{database_filename} written to the same folder as database_filename? It could pollute the directory with unnacessary files. Maybe database_filename/proc{n}.bin would be a better solution?

Updated.

fortran/meshfem3d/shared/command_line_arguments.f90
```diff
@@ -1,74 +0,0 @@
-
```
lsawade, line 1: I guess the `shared` folder could technically be eliminated?

fortran/meshfem2d/shared_par.f90
```diff
@@ -367,7 +367,7 @@ module shared_input_parameters
    logical :: use_binary_for_wavefield_dumps
 
    ! NUMBER_OF_SIMULTANEOUS_RUNS
-   integer :: NUMBER_OF_SIMULTANEOUS_RUNS
+   integer :: NUMBER_OF_SIMULTANEOUS_RUNS = 1
```
lsawade, line 370: Remove? Or, will we support this? With source encoding this seems like a thing of the past

core/specfem/mpi/mpi.hpp
```diff
@@ -109,6 +113,63 @@ class MPI {
     return rank_ == 0;
   }
 
+  /**
```
lsawade, line 117: Should this be the first thing that is tested using MPI? I would suggest a follow up PR, but it could serve as a blueprint

lsawade (approved):

fortran/meshfem2d/shared_par.f90
icui: Source encoding could still use this because we need multiple simulations to avoid a single simulation being too long.

core/specfem/mpi/mpi.hpp
```diff
@@ -109,6 +113,63 @@ class MPI {
...
+    if (lastdot != std::string::npos) {
+      stem = file_name.substr(0, lastdot);
+      ext = file_name.substr(lastdot); // includes the '.'
+    } else {
+      stem = file_name;
+      ext = "";
+    }
+
+    // New scheme: dir/stem/proc_N.ext
+    return dir_path + stem + "/proc_" + proc_str.str() + ext;
```
Rohit-Kakodkar: Can you use boost filesystem to do this. For example, this would not work on windows since windows use `\` delimiters

fortran/meshfem2d/save_databases.f90
```diff
@@ -42,13 +42,51 @@ subroutine save_databases()
...
+            ext = stem(lastdot:len_trim(stem))          ! includes '.'
+            stem = stem(1:lastdot-1)
+         else
+            ext = ''
+         endif
+
+         ! new scheme: dir/stem/proc_N.ext
+         ! rank 0 creates the subdirectory before all ranks write
+         if (iproc == 0) then
+            call execute_command_line('mkdir -p ' // trim(dir_path) // trim(stem), wait=.true.)
```
Rohit-Kakodkar: Try using https://github.com/fortran-lang/stdlib fortran for managing OS calls.

Rohit-Kakodkar (changes requested):

Rohit-Kakodkar (approved):
Thanks for changes!

PR #1725, opened 2026-03-12
Simplified flux scheme yaml node

Rohit-Kakodkar:
## Description

Here is a simple version of the node. We can expand this later.

## Issue Number

## Checklist

### DIFF
--- a/core/specfem/runtime_configuration/flux_schemes.cpp
+++ b/core/specfem/runtime_configuration/flux_schemes.cpp
@@ -8,66 +8,17 @@ specfem::runtime_configuration::flux_schemes::flux_schemes(
     const YAML::Node &Node)
     : flux_schemes_node(Node) {
 
-  // multi-rules in commit 4a27c96cde7e8548556da6caf628eda034fd35b3
-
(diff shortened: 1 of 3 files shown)


core/specfem/runtime_configuration/flux_schemes.cpp
```diff
@@ -8,66 +8,17 @@ specfem::runtime_configuration::flux_schemes::flux_schemes(
     const YAML::Node &Node)
     : flux_schemes_node(Node) {
 
-  // multi-rules in commit 4a27c96cde7e8548556da6caf628eda034fd35b3
```
int-ptr-ptr, line 11: Can we leave this comment in, just to not lose track of it for when we expand the feature-set?

int-ptr-ptr (approved):

core/specfem/runtime_configuration/flux_schemes.cpp
Rohit-Kakodkar: @int-ptr-ptr yeah feel free to take this branch and issue a PR into devel. I think this will require additional changes before the code compiles

PR #1717, opened 2026-03-09
1064 tags - Template point and compute with `specfem::tags::Tags`

icui:
## Description

(Individual PRs have been reviewed)
- Replace point field template parameter with `specfem::tags::Tags`
- Use new template parameter for `stress` and `mass_matrix` computation

## Issue Number

Closes #1064

## Checklist

Please make sure to check developer documentation on specfem docs.

- [ ] I ran the code through pre-commit to check style
- [ ] **THE DOCUMENTATION BUILDS WITHOUT WARNINGS/ERRORS**
- [ ] I have added labels to the PR (see right hand side of the PR page)
- [ ] My code passes all the integration tests
- [ ] I have added sufficient unittests to test my changes
- [ ] I have added/updated documentation for the changes I am proposing
- [ ] I have updated CMakeLists to ensure my code builds
- [ ] My code builds across all platforms

PR #1729, opened 2026-03-12
Revert "1064 tags - Template point and compute with `specfem::tags::Tags`"

lsawade:
Reverts PrincetonUniversity/SPECFEMPP#1717


icui (approved):

Rohit-Kakodkar (commented):
Why?

PR #1733, opened 2026-03-13
Issue 1711 - Support multi-partition adjacency graphs in 2D mesh I/O

### LINKED ISSUE #1711: Compute adjancency graph
Compute adjacency graph within the (mesher?)

- For 2D 
- For 3D

Rohit-Kakodkar:
## Description

Extends the 2D adjacency graph to support multi-partition (distributed) meshes. Previously, the adjacency graph was only usable with a single partition (NPROC=1). This change writes per-partition adjacency data from the Fortran mesher and updates the C++ solver to read and correctly handle cross-partition edges.

### Changes in Meshfem2D

save_databases_adjacency_graph now writes only the adjacency records owned by the current partition (iproc), using local element indices via glob2loc_elmnts. Each record is extended from 4 to 5 integers: (local_elem1, local_elem2, adjacency_type, adjacency_id, neighbor_partition), where neighbor_partition is -1 for intra-partition edges and the remote partition rank for cross-partition edges. Adds a write-count consistency check.
Disables the now-redundant save_databases_interfaces call (MPI interfaces are captured by the adjacency graph).

### Changes in C++ 2D solver

- [x] Added ``local_connections`` and ``mpi_connections`` functions within the mesh struct
- [x] Local connections are stored as adjacency graph
- [x] MPI connections are stored as vector of MPI edge struct

### Tests (to be updated in next PR)

Single-partition (NPROC=1): All neighbor_partition values are -1, so behavior is identical to before — no regression expected.
Multi-partition: Reader correctly parses the 5-integer format, symmetry assertion passes, and assembly/intersection logic operates only on local edges.

## Issue Number

Closes #1711 

## Checklist


lsawade (approved):
In general this looks good to me, let us know if the bug requires larger fix

icui (approved):

Rohit-Kakodkar:
> In general this looks good to me, let us know if the bug requires larger fix

I think I just need to regenerate the databases. I generated some today, but all should be in next week.

PR #1745, opened 2026-03-20
Fix: Clang Memory Limit by splitting updated wavefields and instantiation

lsawade:
## Description

`time_marching.cpp` became way to large resulting in a massive > 1,000,000 line translation unit largely a result of update wavefield macro expansion.  

Splits `update_wavefields` into function defined in the `compute/impl/<func>.tpp` and instantiated in `compute/impl/<func>.cpp` using a simple instantiation macro for compute functions from a list of compute combinations (NGLL, field_type, medium). `update_wavefields` also being instantiated the same way.

## Issue Number

`Intel Devel` CI was failing this should fix it.

## Checklist

### DIFF
--- a/core/specfem/compute/impl/compute_coupling.hpp
+++ b/core/specfem/compute/impl/compute_coupling.hpp
@@ -7,11 +7,11 @@
 namespace specfem::compute::impl {
 
 template <int NGLL, int NQuad_intersection, typename Tags>
-void compute_coupling_weakly_conforming(
+void coupling_core_weakly_conforming(
     const specfem::assembly::assembly<Tags::dimension_tag> &assembly);
 
 template <int NGLL, int NQuad_intersection, typename Tags>
-void compute_coupling_nonconforming(
+void coupling_core_nonconforming(
     const specfem::assembly::assembly<Tags::dimension_tag> &assembly);
 
 /**
@@ -31,16 +31,7 @@ void compute_coupling_nonconforming(
  *
  * @param assembly SPECFEM++ assembly object.
  */
-template <int NGLL, int NQuad_intersection, typename Tags>
+template <int NGLL, typename Tags>
 void compute_coupling(
-    const specfem::assembly::assembly<Tags::dimension_tag> &assembly) {
-  constexpr auto connection_tag = Tags::connection_tag;
-  if constexpr (connection_tag ==
-                specfem::element_connections::type::nonconforming) {
-    compute_coupling_nonconforming<NGLL, NQuad_intersection, Tags>(assembly);
-  } else {
-    compute_coupling_weakly_conforming<NGLL, NQuad_intersection, Tags>(
-        assembly);
-  }
-}
+    const specfem::assembly::assembly<Tags::dimension_tag> &assembly);
 } // namespace specfem::compute::impl
--- a/core/specfem/macros/compute_instantiation_macros.hpp
+++ b/core/specfem/macros/compute_instantiation_macros.hpp
@@ -1,0 +1,89 @@
+#pragma once
+// !! Do NOT include this from specfem/macros.hpp !!
+// Include only in explicit-instantiation .cpp files.
+
+// ── Signature helpers
+// ───────────────────────────────────────────────────────── Callers must have
+// the following using-declarations in scope:
+//   using specfem::element::dimension_tag;
+//   using specfem::element::medium_tag;
+//   using specfem::simulation::field_type;
+
+// void(const assembly &)                  — compute_coupling,
+// divide_mass_matrix
+#define INSTANTIATE_COMPUTE_FUNCTION(FN, NGLL, DIM, WF, MED)                   \
+  template void                                                                \
+  specfem::compute::impl::FN<NGLL, specfem::tags::Tags<DIM, WF, MED> >(        \
+      const specfem::assembly::assembly<DIM> &);
+
+// void(assembly &, const int)             — compute_source_interaction
+#define INSTANTIATE_COMPUTE_FUNCTION_VOID_NONCONST_INT(FN, NGLL, DIM, WF, MED) \
+  template void                                                                \
+  specfem::compute::impl::FN<NGLL, specfem::tags::Tags<DIM, WF, MED> >(        \
+      specfem::assembly::assembly<DIM> &, const int);
+
+// int(const assembly &, const int)        — compute_stiffness_interaction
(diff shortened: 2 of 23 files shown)


lsawade:
@int-ptr-ptr adding you because this is relevant for your coupling kernels.

Rohit-Kakodkar (approved):

core/specfem/compute/impl/compute_coupling.hpp
```diff
@@ -7,11 +7,11 @@
 namespace specfem::compute::impl {
 
 template <int NGLL, int NQuad_intersection, typename Tags>
-void compute_coupling_weakly_conforming(
+void coupling_core_weakly_conforming(
     const specfem::assembly::assembly<Tags::dimension_tag> &assembly);
```
int-ptr-ptr, line 11: Weakly conforming does not have any `NQuad_intersection` dependence. It looks like that got added in with the <..., Tags> change, namely commit a51f4ea9daa7eef0398e7b39145c2d65da2f76dd.

Surprisingly, before then, NGLL was not a template param in `compute_coupling_weakly_conforming`. This might be a reason to split `weakly-conforming` couplings and `nonconforming` couplings on the `compute`-side, on top of the `assembly`-side split.

core/specfem/macros/compute_instantiation_macros.hpp
```diff
@@ -0,0 +1,89 @@
...
+//   using specfem::element::dimension_tag;
+//   using specfem::element::medium_tag;
+//   using specfem::simulation::field_type;
+
+// void(const assembly &)                  — compute_coupling,
+// divide_mass_matrix
+#define INSTANTIATE_COMPUTE_FUNCTION(FN, NGLL, DIM, WF, MED)                   \
+  template void                                                                \
+  specfem::compute::impl::FN<NGLL, specfem::tags::Tags<DIM, WF, MED> >(        \
+      const specfem::assembly::assembly<DIM> &);
```
int-ptr-ptr: This will have to change pretty soon, since I plan on looking at different `Nquad_intersection`s very soon, which came about after talking to Jeroen on Thursday. That means that (unless we're fine with zero-padding intersection quadratures for the forseeable future) we would need `impl::compute_coupling<NGLL, NQUAD_INTERSECTION, specfem::tags::Tags<...>>`, which messes up this whole macro setup, especially considering only `compute_coupling_nonconforming` needs both `NGLL` and `Nquad_intersection`.

int-ptr-ptr (commented):
To me, it feels like this should be transposed. Instead of
```
def INST_MACRO(MACRO):
  MACRO(ARGSET1) // dim2, ngll5, forward
  MACRO(ARGSET2)
  MACRO(ARGSET3) // dim2, ngll5, backward
  ...
```
it seems more reasonable to do
```
def INST_ARGSET(ARGSET):
  MACRO1(ARGSET) // compute coupling
  MACRO2(ARGSET) // source interaction
  MACRO3(ARGSET) // compute stiffness
  ...
```
since a missing unit for a given `(NGLL, dimension_tag, field_type)` tuple will cause the whole thing to grind to a halt, at least outside of unit tests. I get that the translation units line up with the `compute::impl` file-structure, but to me:
- the purpose of the `.cpp` files are not obvious, since they snake to `macros` and back to `.tpp`. The triple macro expansion feels very opaque at first.
- `INST_ARGSET` lines up closer to `update_wavefields`. If I'm not mistaken on how translation units work, you could drop the `impl/*.cpp` files and use `update_wavefields` as the instantiator, dropping the FOR_EACH_IN_PRODUCT macros. Then, instantiate `update_wavefields` with different argsets in each translation unit.
- having different translation units for different argset groupings would make your future plans of partial / specialized compiles much easier, since a dim2 compile can just include the dim2 `.cpp` files, etc.

  Even before then, if you're implementing a dim3 change, or fixing an acoustic bug or something, then only that unit needs recompiling. Unless I'm wrong about translation units again, that should speed up partial rebuilds.

lsawade:
@int-ptr-ptr 

I think you misunderstand what's happening when compiling. Currently, everything is compiled when `time_marching.cpp` instantiates with `ngll` and `field`. That means when the preprocessor/compiler hits that instantiation, it will put every single definition that is required for `time_marching.cpp` into `/tmp/time_marching_#someid.cpp`. That means all macros for all media and all combinations are currently put into a single translation unit, resulting in a single massive `.cpp` file to be compiled. This means all definitions for all the different applicable physics, etc., need to be in a single compilable file.

If I do what you suggest and let `update_wavefields` instantiate the different functions, the only thing that happens is that you split `time_marching.cpp`, since `time_marching.cpp` now only needs the `update_wavefields.hpp` header because `update_wavefields.cpp` is already compiled. However, each `update_wavefield` will still contain ALL definitions for all physics (it needs to include all `.tpp` files). I fear that this would only be a temporary solution, because the moment we expand the macro outer product — in particular with new media in 3D (@maxlchien is adding 3D Cosserat right now, e.g.) — we will hit this same issue because `/tmp/update_wavefields_#someid.cpp` will become too large for Clang to compile. In this PR, `update_wavefields` only needs the `impl/compute_x.hpp` header without the full definition of the functions. That is the point: the temporary translation unit becomes smaller.

Then, the last point, I totally get and you are correct, but you do understand what that means right? That means that we would have one file for each valid combination. ngll, field, dim, medium, property etc. that would be a lot of files to maintain. And the gain here may be small because for each file then ALL necessary template definitions and headers need to parsed separately for each translation unit which becomes a huge burden on preprocessor. If you meant each of them should have the valid `compute_x_valid_compute_combination.cpp` in terms of `ngll`, `field` etc. I'm also trying very hard to strike a good middle between memory usage and file maintenance. If we approached it that way I could easily think that we could precompiled the header compute_x.hpp which would the be included by the `compute_x_<combination>.cpp` removing some overhead. But for this there is another overhead because now every translation unit is waiting for that header to be compiled (translated to binary), potentially stalling parallel processes. I recommend: [Youtube Link: Why Is My C++ Build So Slow? Compilation Profilin...](https://www.youtube.com/watch?v=Oih3K-3eZ4Y).

For the macro and calling/structure, I think you are making good points, but that may change anyway in a couple of weeks because we have a prototype to eliminate the macros. I would prefer to leave a refactor of that to a later point.

lsawade:
@int-ptr-ptr follow-up. You gave me an idea. we could have compute_x.cpp.in and do the same thing using cmake [Claude made prototype](https://claude.ai/share/4afe1e63-8297-41df-b462-0bf8d029d740), but for this we really would need a precompiled header which ***IS*** ***~not~*** a big deal. The header gets too large.

icui (approved):

Issue #1702, opened 2026-03-04
Source / Receiver Location Across MPI Slices


icui:
**Files to modify:** `core/specfem/assembly/sources/` (dim2 and dim3)

**Background:**
Each source is found by searching all elements on each rank. The rank with the smallest
misfit "wins". All ranks must know the owner rank of each source.

**Algorithm** (from `mpi_ref/specfem3D/locate_MPI_slice.f90`):
1. Each rank computes local best-fit `(dist, ispec, xi, eta, gamma)` per source
2. `allreduce(min, dist_per_source[nsources])` → global minimum distances
3. Rank whose `local_dist[isrc] == global_min_dist[isrc]` is the owner; it sets `islice_selected_source[isrc] = myrank`
4. `bcast(islice_selected_source, nsources)` from the owning rank (or use `allreduce(max, islice_array)`)
5. During time stepping: guard source injection with `MPI::get_rank() == islice_selected_source[isrc]`

**Edge cases:**
- Multiple sources may be owned by different ranks
- If two ranks tie on distance, use the higher rank (matches reference behavior)
- Print source summary (location, depth, distance) only from rank 0

**Ref:** `mpi_ref/specfem3D/locate_source.F90`, `mpi_ref/specfem3D/locate_MPI_slice.f90`

Rohit-Kakodkar:
See ArborX for doing source location. It's a parallel BVM library that would help. Internally build using kokkos views

PR #1757, opened 2026-03-26
Issue 1702 - Locate sources / stations with MPI

icui:
## Description

Each process calculates the element closesest to a source / receiver, then performs a global reduction. More optimzied algorithms will be PRed separately.

## Issue Number

Adds to #1702

## Checklist


core/specfem/assembly/assembly/dim2/assembly.cpp
Rohit-Kakodkar, line 96: Should this function be called only by root?

core/specfem/assembly/assembly/dim2/assembly.cpp
```diff
@@ -149,5 +150,40 @@ specfem::assembly::assembly<specfem::element::dimension_tag::dim2>::print()
     message << "   Elastic media will simulate P-SV polarized waves\n";
   }
 
+#ifdef SPECFEM_ENABLE_MPI
+  {
+    const auto &src_islice = this->sources.get_islice();
+    if (!src_islice.empty()) {
```
Rohit-Kakodkar: Why do sources and receivers have a different pattern for generating the message?

core/specfem/assembly/assembly/dim2/assembly.cpp
```diff
@@ -149,5 +150,40 @@ specfem::assembly::assembly<specfem::element::dimension_tag::dim2>::print()
...
+      std::map<int, std::vector<int> > sources_per_rank;
+      for (int i = 0; i < static_cast<int>(src_islice.size()); ++i)
+        sources_per_rank[src_islice[i]].push_back(i);
+      message << "Sources per MPI rank:\n";
```
Rohit-Kakodkar: Seems like every MPI process is generating its own separate message? Is that what we want?

core/specfem/assembly/receivers/dim2/receivers.cpp
```diff
@@ -63,34 +64,67 @@ specfem::assembly::receivers<specfem::element::dimension_tag::dim2>::receivers(
...
+  }
+
+  std::vector<type_real> global_dists = local_dists;
+  SPECFEM_MPI_SAFECALL(MPI_Allreduce(
+      MPI_IN_PLACE, global_dists.data(), nreceivers,
+      specfem::MPI::datatype<type_real>(), MPI_MIN, MPI_COMM_WORLD));
+
+  std::vector<int> islice_selected(nreceivers, -1);
+  for (int i = 0; i < nreceivers; ++i) {
+    if (local_dists[i] <= global_dists[i])
```
Rohit-Kakodkar: I assume what you are doing here is figuring out if  my distance is the smallest distance. Should you then be checking that? 

What if MPI all reduce call had a numerical error that resulted in global_distance[i] = local_dist[i] - \epsilon ?

Rohit-Kakodkar: Is this going to work for edge cases? See: https://github.com/SPECFEM/specfem3d/blob/21e40fd1f473caaa0771fec16d39c3cfab1d05e2/src/specfem3D/locate_MPI_slice.f90#L112

@lsawade I feel you know more about this.

Rohit-Kakodkar (changes requested):
See comments.

We should build tests for this before we merge. I feel this would be easy to get wrong.

core/specfem/algorithms/locate_point_mpi_slice.hpp
```diff
@@ -0,0 +1,149 @@
...
+ * @param mesh    This rank's local mesh partition.
+ * @return { local_coords, owning_ranks }
+ *         local_coords[i] is valid (ispec >= 0) only when owning_ranks[i]
+ *         equals this rank's MPI rank; all other entries have ispec = -1.
+ * @throws std::runtime_error if any point cannot be located on any rank.
+ */
+template <specfem::element::dimension_tag DimensionTag>
+std::pair<std::vector<specfem::point::local_coordinates<DimensionTag> >,
+          std::vector<int> >
+locate_point_mpi_slice(
```
Rohit-Kakodkar: Should you be calling this function within ``locate_point``. Make this function impl since it is not to be exposed.

core/specfem/assembly/assembly/dim2/assembly.cpp
Rohit-Kakodkar: See the comment above.

core/specfem/assembly/assembly/dim2/assembly.cpp
Rohit-Kakodkar: see comment abpve

core/specfem/assembly/receivers/dim2/receivers.cpp
```diff
@@ -63,34 +63,38 @@ specfem::assembly::receivers<specfem::element::dimension_tag::dim2>::receivers(
...
+  std::vector<specfem::point::global_coordinates<
+      specfem::element::dimension_tag::dim2> >
+      gcoords;
+  gcoords.reserve(nreceivers);
+  for (int ireceiver = 0; ireceiver < nreceivers; ++ireceiver)
+    gcoords.push_back(
+        { receivers[ireceiver]->get_x(), receivers[ireceiver]->get_z() });
+
+  auto [local_coords, islice_selected] =
+      specfem::algorithms::locate_point_mpi_slice(gcoords, mesh);
```
Rohit-Kakodkar: It would be nice if ``locate_point`` would return the ``local_coordinates`` when the point is in current mpi slice and a null type when it is not.

This interface would be very clear

```
for (auto receiver : receivers) {
    auto gcoord = {receiver.get_x(), receiver.get_z()};
    // The if condition is only satisfied if point is within this domain. Similar to YAML nodes. 
    if (local_coordinates local_coord = specfem::algorithms::locate_point(...)) {
        // do logic here
    }
}
```

core/specfem/mpi/mpi.hpp
```diff
@@ -234,23 +181,15 @@ class MPI {
...
   friend void specfem::program::abort(const std::string &, int, const int,
                                       const char *);
 };
 
 } // namespace specfem
+
+#ifndef SPECFEM_ENABLE_MPI
+#define SPECFEM_MPI_SAFECALL(call) ((void)0)
+#else
+#define SPECFEM_MPI_SAFECALL(call) (specfem::MPI::check_context(), (call))
```
Rohit-Kakodkar:
```suggestion
#define SPECFEM_MPI_SAFECALL(call) do { \
    specfem::MPI::check_context(); \
    int e = call; \ 
    if (e != MPI_SUCCESS) {                           \
        char err[MPI_MAX_ERROR_STRING];                 \
        int len;                                        \
        MPI_Error_string(e, err, &len);                 \
        fprintf(stderr, "MPI error %s:%d: %s\n",        \
                __FILE__, __LINE__, err);               \
        MPI_Abort(MPI_COMM_WORLD, e);                   \
    } \
} while(0)
```

Rohit-Kakodkar (changes requested):

core/specfem/algorithms/locate_point_mpi_slice.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <cmath>
+#include <limits>
+#include <stdexcept>
+#include <utility>
+#include <vector>
+
+namespace specfem {
+namespace algorithms {
+
+namespace detail {
```
lsawade: we use impl

core/specfem/assembly/assembly/dim3/assembly.cpp
```diff
@@ -120,5 +121,41 @@ specfem::assembly::assembly<specfem::element::dimension_tag::dim3>::print()
...
+        sources_per_rank[src_islice[i]].push_back(i);
+      message << "Sources per MPI rank:\n";
+      for (const auto &[rank, indices] : sources_per_rank) {
+        message << "  Rank " << rank << " :";
+        for (int idx : indices)
+          message << " " << idx;
+        message << "\n";
+      }
+    }
+    if (this->receivers.size() > 0) {
```
lsawade: Same here to @Rohit-Kakodkar 's comment

core/specfem/assembly/sources/impl/locate_sources.tpp
```diff
@@ -6,46 +6,44 @@
...
+  std::vector<specfem::point::global_coordinates<DimensionTag>> coords;
+  coords.reserve(nsources);
+  for (int isrc = 0; isrc < nsources; ++isrc)
+    coords.push_back(sources[isrc]->get_global_coordinates());
+
+  // Locate all sources across MPI partitions with inside-preference.
+  // The function prefers elements where xi/eta/gamma ∈ [-1,1] and falls back
+  // to minimum Cartesian distance when no rank owns an inside element.
+  auto [lcoords, islice_selected] =
+      specfem::algorithms::locate_point_mpi_slice(coords, mesh);
```
lsawade: this name should still be `locate_point(...)` and then with the algorithm should choose which one.

core/specfem/assembly/receivers/dim2/receivers.cpp
lsawade: I think the receiver point location could be strongly simplified, if 2D and 3D receivers mimic the `source` and have `get_global_coordinates(...)` method. Then there's no need for the auto gcoord either.

lsawade: ... of course `set_local_coordinates()` should be there too

core/specfem/assembly/sources/impl/locate_sources.tpp
```diff
@@ -6,46 +6,44 @@
...
-      throw std::runtime_error("Source is outside of the domain");
-    }
-
-    // Giving the local coordinates and global element index to the source
-    source->set_local_coordinates(lcoord);
-
-    // Given the spectral element index provide the medium tag
-    source->set_medium_tag(element_types.get_medium_tag(lcoord.ispec));
   }
 }
```
lsawade, line 49:
```suggestion
// Function that can be used for both sources and receivers.
template <specfem::element::dimension_tag DimensionTag, typename PointLocationType>
void specfem::assembly::sources_impl::locate_sources(
    const specfem::assembly::element_types<DimensionTag> &element_types,
    const specfem::assembly::mesh<DimensionTag> &mesh,
    std::vector<std::shared_ptr<PointLocationType> >
        &points) {

  static_assert("make sure PointLocationType matches element_types, and mesh", ...)

  const int npoints = static_cast<int>(points.size());
  const int myrank = specfem::MPI::get_rank();

  // Collect global coordinates for all sources.
  std::vector<specfem::point::global_coordinates<DimensionTag>> coords;
  coords.reserve(nsources);
  for (int ipt = 0; ipt < npoints; ++ipt)
    coords.push_back(points[ipt]->get_global_coordinates());

  // Locate all points across MPI partitions with inside-preference.
  // The function prefers elements where xi/eta/gamma ∈ [-1,1] and falls back
  // to minimum Cartesian distance when no rank owns an inside element.
  auto [lcoords, islice_selected] =
      specfem::algorithms::locate_point_mpi_slice(coords, mesh);

  // Assign local coordinates and medium tags to each source.
  // Only the owning rank sets valid coordinates; all others receive ispec = -1
  // so that downstream filtering skips them.
  for (int ipt = 0; ipt < nsources; ++ipt) {
    points[ipt]->set_islice(islice_selected[ipt]);
    if (islice_selected[ipt] == myrank) {
      const auto &lcoord = lcoords[ipt];
      points[ipt]->set_local_coordinates(lcoord);
      points[ipt]->set_medium_tag(element_types.get_medium_tag(lcoord.ispec));
    } else {
      specfem::point::local_coordinates<DimensionTag> invalid;
      invalid.ispec = -1;
      points[ipt]->set_local_coordinates(invalid);
    }
  }
}
```

lsawade (changes requested):

core/specfem/assembly/assembly/dim2/assembly.cpp
icui: By default logger only prints from rank 0, for debugging I think each proc generating its own message makes sense because you can check if they are synced correctly.

core/specfem/assembly/assembly/dim2/assembly.cpp
icui: The main difference is that there is a `StationIterator` and `receivers` gathers all names, islices etc into vectors.

core/specfem/assembly/receivers/dim2/receivers.cpp
icui: `islice` for receiver is still needed.

core/specfem/algorithms/locate_point/dim2/locate_point_impl.cpp
```diff
@@ -367,6 +370,130 @@ locate_point_core(
...
+locate_point(
+    const specfem::point::global_coordinates<
+        specfem::element::dimension_tag::dim2> &coordinates,
+    const specfem::assembly::mesh<specfem::element::dimension_tag::dim2>
+        &mesh) {
+  return locate_point_core(mesh.graph(), coordinates, mesh.h_coord,
+                           mesh.h_control_node_coord, mesh.ngnod);
+}
+
+specfem::point::global_coordinates<specfem::element::dimension_tag::dim2>
```
lsawade, line 385: This function name does not make sense. It returns global coordinates from a local coordinate. It maps.

core/specfem/algorithms/locate_point/dim2/locate_point_impl.cpp
```diff
@@ -367,6 +370,130 @@ locate_point_core(
...
+      point::global_coordinates<specfem::element::dimension_tag::dim2> *,
+      Kokkos::HostSpace>
+      coorg("coorg", ngnod);
+
+  for (int i = 0; i < ngnod; i++) {
+    coorg(i).x = mesh.h_control_node_coord(0, ispec, i);
+    coorg(i).z = mesh.h_control_node_coord(1, ispec, i);
+  }
+
+  return jacobian::compute_locations(coorg, ngnod, xi, gamma);
```
lsawade, line 406: Similarly this function `compute_location(s)`. This is `local_to_global_mapping(xi, gamma, control_nodes)`. See equation 20 in Komatitsch & Tromp 1999.

core/specfem/algorithms/locate_point/dim2/locate_point_impl.cpp
lsawade: Just adding a comment 
```suggestion
// TODO ( Any : Add KDTree ) This is right where a KDTree would slot in.
std::tie(ix_guess, iz_guess, ispec_guess) =
rough_location(coordinates, global_coordinates);
```

core/specfem/algorithms/locate_point/dim3/locate_point_impl.cpp
lsawade, line 273: Just adding a comment 
```suggestion
// TODO ( Any : Add KDTree ) This is right where a KDTree would slot in.
std::tie(ispec_guess, ix_guess, iy_guess, iz_guess) =
rough_location(coordinates, global_coordinates);
```

lsawade (approved):
I have a couple of unrelated comments on this. I think function naming on some of these is misleading, but that should be addressed separately.

core/specfem/io/mesh/impl/fortran/dim2/mesh.cpp
```diff
@@ -79,19 +79,18 @@ specfem::io::read_2d_mesh(
 
 #if defined(SPECFEM_ENABLE_MPI)
   int nspec_all = mesh.parameters.nspec;
+  int nelem_acforcing_all = mesh.parameters.nelem_acforcing;
+  int nelem_acoustic_surface_all = mesh.parameters.nelem_acoustic_surface;
```
Rohit-Kakodkar: This does not need to be in an mpi guard anymore.

Rohit-Kakodkar: This can also be replaced with an inplace reduce.

core/specfem/assembly/assembly/dim2/assembly.cpp
```diff
@@ -144,5 +145,40 @@ specfem::assembly::assembly<specfem::element::dimension_tag::dim2>::print()
     message << "   Elastic media will simulate P-SV polarized waves\n";
   }
 
+#ifdef SPECFEM_ENABLE_MPI
+  {
```
Rohit-Kakodkar: Should this be printed within the source struct?

core/specfem/algorithms/locate_point.hpp
```diff
@@ -1,139 +1,151 @@
...
+  SPECFEM_MPI_SAFECALL(MPI_Allreduce(MPI_IN_PLACE, global_priority.data(),
+                                     npoints, SPECFEM_MPI_TYPE_REAL, MPI_MIN,
+                                     MPI_COMM_WORLD));
+
+  // Step 2: each rank claims points whose local priority matches the global
+  // minimum; allreduce(max) resolves ties by assigning to the highest rank
+  // (matching legacy SPECFEM behaviour).
+  std::vector<int> islice_selected(npoints, -1);
+  for (int i = 0; i < npoints; ++i) {
+    if (local_priority[i] <= global_priority[i])
```
Rohit-Kakodkar: I think you should be using ``is_close`` here. Incase there an a numerical error when doing all reduce in global priority, this might not pass.

core/specfem/assembly/sources/impl/source_medium.hpp
```diff
@@ -365,6 +365,12 @@ sort_sources_per_medium(
 
   for (int isource = 0; isource < (int)sources.size(); isource++) {
     const auto &source = sources[isource];
+
+    // In MPI builds a source not owned by this rank has ispec = -1.
+    // Skip it here so it is never added to any medium's source list.
+    if (source->get_local_coordinates().ispec < 0) {
+      continue;
+    }
```
Rohit-Kakodkar: This should use the souce partition (islice) to check against my_rank since thats more clear.

Rohit-Kakodkar (changes requested):
Minor comments.

core/specfem/algorithms/locate_point.hpp
icui: I don't think MPI_Allreduce can change the value so I could use `==` here. Using `is_close` may cause issue when two distances are actually close.

core/specfem/assembly/assembly/dim3/assembly.cpp
Rohit-Kakodkar, line 1: This is not you. But I realize now that our messages get computed on all processes and only printed on the root. That's a wasted computation; we could wrap them within the ``SPECFEM_ON_ROOT`` macro.

Rohit-Kakodkar (approved):
This is great!

PR #1766, opened 2026-04-01
Issue 1711 - MPI database tests

Rohit-Kakodkar:
## Description

MPI tests for reading the MPI connections

- Update MPI database to make it smaller
- Updates assembly adjacency graph for compilation
- Adds an MPI mesh reader test
## Issue Number

## Checklist


tests/unit-tests/mesh/dim3/adjacency_graph.cpp
```diff
@@ -133,3 +134,212 @@ TEST_P(Mesh3DTest, AdjacencyGraph) {
...
+              << ", anchor " << static_cast<int>(expected_mpi_conn.edge2.anchor)
+              << ")\n";
+          msg << "  No matching MPI connection found in adjacency graph.\n";
+          ADD_FAILURE() << msg.str();
+        });
+      }
+
+      SPECFEM_MPI_SAFECALL(MPI_Barrier(MPI_COMM_WORLD));
+    }
+    SPECFEM_MPI_ON_ROOT({
```
lsawade, line 260: Is this the way to do it? I mean is this teh way MPI is done with google test?

lsawade: Found this https://github.com/DLR-SC/googletest_mpi , but they only have a couple of devs albeit at the DLR (German NASA).

lsawade (approved):

icui (approved):

tests/unit-tests/mesh/dim3/adjacency_graph.cpp
Rohit-Kakodkar: The ``SUCCEED`` will cause only the root to print this message. 

This current test would, for example, only ever fail on root. I could make it fail on all procs above, but then the error will be printed 4 times. Either is okay, I chose to not be too verbose I guess.

Issue #1719, opened 2026-03-10
Material / Interface iterator macros are too expensive for clang


Rohit-Kakodkar:
## Task description

This is a limitation of our ``FOR_EACH`` infrastructure. The iterators are memory-intensive for compilation. 

## Problem

Currently, all valid sequences are given as individual sequences, i.e. 
```
#define EDGES                                                                  \
  ((DIMENSION_TAG_DIM2, CONNECTION_TAG_WEAKLY_CONFORMING,                      \
    INTERFACE_TAG_ELASTIC_ACOUSTIC, BOUNDARY_TAG_NONE))(                       \
      (DIMENSION_TAG_DIM2, CONNECTION_TAG_WEAKLY_CONFORMING,                   \
       INTERFACE_TAG_ELASTIC_ACOUSTIC, BOUNDARY_TAG_STACEY))(                  \
      ...
``` 

We then compare each tag within the ``FOR_EACH`` target sequence to above to check if we find a match. However, this generates a lot of macro code, primarily through the Boost Preprocessor Library. This leads to clang compiler running out of memory. On my machine :

```
clang: error: unable to execute command: Segmentation fault: 11
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin25.3.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /var/folders/nf/pg3tk3lj1h7grxnbkbkp23dm0000gn/T/time_marching-fb171b.cpp
clang: note: diagnostic msg: /var/folders/nf/pg3tk3lj1h7grxnbkbkp23dm0000gn/T/time_marching-fb171b.sh
clang: note: diagnostic msg: Crash backtrace is located in
clang: note: diagnostic msg: /Users/rohit-kakodkar/Library/Logs/DiagnosticReports/clang_<YYYY-MM-DD-HHMMSS>_<hostname>.crash
clang: note: diagnostic msg: (choose the .crash file that corresponds to your crash)
clang: note: diagnostic msg: 

********************
[...]

PR #1800, opened 2026-04-15
Issue 1719 -  `tag_dispatch` for value_containers.

icui:
## Description

Updates `tag_dispatch` for `value_containers`, `assembly::kernels`, `assembly::properties` and merge 2D / 3D implementations when possible.

## Issue Number

Adds to #1719

## Checklist


core/specfem/assembly/impl/value_containers.hpp
```diff
@@ -1,19 +1,134 @@
...
+ * @tparam ContainerSetsType  A ContainerSets<DimensionTag> specialisation
+ * providing
+ *                            @c dimension_tag and @c combinations.
+ * @tparam containers_type    3-parameter class template
+ *                            (dimension_tag, medium_tag, property_tag).
  */
 template <
-    specfem::element::dimension_tag,
+    typename ContainerSetsType,
     template <specfem::element::dimension_tag, specfem::element::medium_tag,
```
Rohit-Kakodkar, line 47: Should the container type be templated on tags type?

core/specfem/assembly/kernels.cpp
```diff
@@ -0,0 +1,39 @@
...
+  const int nspec = element_types.nspec;
+
+  this->nspec = nspec;
+  this->element_grid = grid;
+
+  this->property_index_mapping =
+      Kokkos::View<int *, Kokkos::DefaultExecutionSpace>(
+          "specfem::assembly::kernels::property_index_mapping", nspec);
+  this->h_property_index_mapping =
+      Kokkos::create_mirror_view(this->property_index_mapping);
```
Rohit-Kakodkar: This should be initialized using member initializer list i.e. ":" syntax.

core/specfem/assembly/kernels.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+  specfem::tag_dispatch::for_each(this->combinations, [&]<typename TagsType>() {
+    this->value
+        .template get<TagsType>() = specfem::assembly::impl::domain_kernels<
+        TagsType::dimension_tag, TagsType::medium_tag, TagsType::property_tag>(
+        element_types.get_elements_on_host(TagsType::medium_tag,
+                                           TagsType::property_tag,
+                                           TagsType::attenuation_tag),
+        grid, this->h_property_index_mapping);
+  });
```
Rohit-Kakodkar: Can use initializer functor call here right instead of for each directly. Then you could also move this to initializer list.

Rohit-Kakodkar (changes requested):

Rohit-Kakodkar:
We had spoken about this at some point, but here is a clarification on how initializer list operates:

The C++ standard **guarantees** that members are initialized in the **order they are declared in the class**, regardless of the order you write them in the initializer list.

```cpp
class Foo {
    int x;  // initialized first
    int y;  // initialized second
public:
    Foo() : y(5), x(y) {}  // ⚠️ DANGER: x is initialized first, y is uninitialized!
};
```

Even though `y(5)` appears first in the list, `x` is initialized first because it's declared first in the class. So `x(y)` reads an uninitialized `y` — undefined behavior.

**The rules:**
1. **Base classes** are initialized first, in the order they appear in the inheritance list
2. **Member variables** are then initialized in declaration order in the class body
3. The order in the initializer list is **irrelevant** to execution order

**Practical takeaway:** Always write your initializer list in the same order as your declarations. Most compilers (GCC, Clang) will even warn you with `-Wall` or `-Wreorder` if the orders don't match, precisely because mismatches are a common source of subtle bugs.

Rohit-Kakodkar (approved):

lsawade (approved):

PR #1828, opened 2026-05-06
Issue 1699 - Communication Buffers

Rohit-Kakodkar:
## Description

- [x] Implements communication buffers
- [x] ~Implements message tag generators~
- [x] Updates tests
- [x] Fixes a bug in MPI communication pattern construction that could've lead to deadlocks

## Issue Number

## Checklist


lsawade (commented):
Looks good to me. Copilot has found an issue with the expression for tag generation.

Can you create a test for the tagging?

Rohit-Kakodkar:
> Looks good to me. Copilot has found an issue with the expression for tag generation.
> 
> Can you create a test for the tagging?

@lsawade While looking to address Copilot comments, I realized the generator is an overkill (not needed). I was initially under the impression that all ranks need to have unique tags, but that's not true. The tags are already distinguished by the sender rank.  So only multiple messages from the same sender need to be unique in the tag, which does not require a generator. So I removed the generator it in the recent commit.

icui (approved):

lsawade (approved):

PR #1833, opened 2026-05-07
759 - Add seismogram write tests

### LINKED ISSUE #759: Write SAC seismograms
## Task description

SAC is still a widely used standard for processing seismic traces as the file containes both metadata and data.

- [ ] Copy writer from specfem3d_globe: [github sf3dg sac writer](https://github.com/SPECFEM/specfem3d_globe/blob/master/src/specfem3D/write_output_SAC.f90)
- [ ] Translate it to C++ and adapting it to the IO scheme in SF++

## Acceptance criteria
- [ ] Can create (correct) readable `.sac` traces.

icui:
## Description

Test seismogram write for SAC and ASCII.

## Issue Number

Closes #759

## Checklist


Rohit-Kakodkar (changes requested):
I am worried that these tests would be hard to maintain over time. There are about 1100 lines of generated code, which are hard to review as is. Maybe it would be much simpler to check the sac output against a reference file. I dont know much about sac so I might be wrong.

tests/unit-tests/io/seismogram/seismogram_writer_tests.cpp
```diff
@@ -0,0 +1,1183 @@
...
+  }
+}
+
+// ===========================================================================
+// 5. SAC helper function tests
+// ===========================================================================
+
+TEST(SACHelpers, ExtractChannelCodeNormal) {
+  EXPECT_EQ(specfem::io::impl::extract_channel_code_from_filename(
+                "SY.ANMO.S3.BXZ.semd"),
```
lsawade: Didn't you say that we should not have a reader as per previous discussions? If no, why would we need this function? 

If yes, I think think there should be a couple more tests, e.g. "N...BHX.sac", "N.S..BHX.semd.sac" in case someone fiddles with that function, tests should be foolproof.

lsawade (changes requested):
The tests are quite long, I would split them at least into `seismogram_writer_tests/sac.cpp`, `seismogram_writer_tests/ascii.cpp`, etc. so that they can be maintained separately.

The other comment above is mostly related to a discussion we hade before. If you need the channel code from the filename at write time instead, we should change the way the sac writer gets the channel and not extract it again from a generated filename

tests/unit-tests/io/seismogram/seismogram_writer_tests.cpp
```diff
@@ -0,0 +1,250 @@
+#include "../../test_macros.hpp"
```
lsawade, line 1: How are the SACs being generated?

Rohit-Kakodkar (approved):

lsawade (approved):

Issue #1819, opened 2026-04-29
Messages need to be converted to include MPI information.


lsawade:
## Task description

- Wrap print messages in SPECFEM_MPI_ON_ROOT
- Update messages to collective information.

## Acceptance criteria
- [ ] Messages are correct across the mpi partitions.

PR #1834, opened 2026-05-07
1819 - Avoid unnecessary print() calls when not printed.

icui:
## Description

Add overloads to `specfem::Logger::debug()` and `specfem::Logger::info()` to take a lambda function that is called only when message is to be printed.

## Issue Number

Adds to #1819

## Checklist


lsawade (commented):
LGTM, but double check copilots comment.

Can you make a table with rows of log levels, columns with serial, mpi root to file, mpi other ranks to file, mpi root to stdout, mpi other ranks to stdout.

and then put 0/1 or true/false's for expected output? I'm trying to follow when what is printed and it's becoming difficult (not because of this PR, but because the logger impl is large)

lsawade (approved):

Rohit-Kakodkar (approved):
Looks great!

PR #1847, opened 2026-05-11
Locate point on face

### LINKED ISSUE #1845: `locate_point_on_face` algorithm
## Task

- add `locate_point_on_face` to dim3 algorithm code
- add test similar to `locate_point_on_edge`

## Acceptance Criteria

- test passes

int-ptr-ptr:
## Description

- adds `locate_point_on_face`
- adds `locate_point_on_entity<dimenstion_tag>`, which delegates to `locate_point_on_edge` or `locate_point_on_face`
- creates test for `locate_point_on_face` (only added to serial)

## Issue Number

#1845

## Checklist


core/specfem/algorithms/locate_point/dim2/locate_point_impl.cpp
```diff
@@ -362,9 +362,9 @@ locate_point_core(
   int ix_guess, iz_guess, ispec_guess;
 
   std::tie(ix_guess, iz_guess, ispec_guess) =
-// TODO ( Any : Add KDTree ) This is right where a KDTree would slot in.
-std::tie(ix_guess, iz_guess, ispec_guess) =
-rough_location(coordinates, global_coordinates);
+      // TODO ( Any : Add KDTree ) This is right where a KDTree would slot in.
```
Rohit-Kakodkar, line 366: Seems like there is a redundant assignment happening here. Probably happened during merge conflict at some point earlier

core/specfem/algorithms/locate_point/locate_point_impl.hpp
```diff
@@ -11,6 +11,16 @@
 // Implementation details exposed for testing
 namespace specfem::algorithms::locate_point_impl {
 
+/**
+ * @brief A type for the local coordinates along a given face (codimension 1)
+ *
+ * @tparam dimension_tag the dimension of the element.
+ */
+template <specfem::element::dimension_tag dimension_tag>
+using facial_coordinate_type =
```
Rohit-Kakodkar: I think this should be removed, since this design exposes this impl type to the outside modules. You could just directly replace the individual types where they are used.

core/specfem/algorithms/locate_point.hpp
```diff
@@ -157,5 +157,91 @@ locate_point(
...
+ *
+ * @param coordinates Local coordinate along face
+ * @param mesh 2D spectral element mesh
+ * @param ispec Element index whose local coordinates to find
+ * @param constraint Edge to compute for
+ * @return Global coordinates of the point
+ */
+specfem::point::global_coordinates<specfem::element::dimension_tag::dim3>
+locate_point_on_face(
+    const specfem::algorithms::locate_point_impl::facial_coordinate_type<
```
Rohit-Kakodkar: Here the interface is opaque since I am not sure what type is the first argument. Also is it better to implement this as just an overload of the ``locate_point`` with another constraint argument. You can then pass the ``local_coordinate`` and we can check that the constraint is met on the local coordinate.

Rohit-Kakodkar (changes requested):
I am mostly fine with this. The core algorithm looks good. 

The design needs to change. I am now wondering whether all these functions could have been implemented as overloads of ``locate_point`` with the mesh entity constraint as the last argument.

We can change the interface so that locate_point_on_entity is an overloaded version of locate_point. Can you remove the impl typedef and just expose it directly? See the 2nd comment

lsawade (dismissed):
@int-ptr-ptr I'm removing @icui and me from the reviewers, please rerequest once fixes are in. Let us know when things are not clear.

lsawade:
did not mean to approve...

core/specfem/algorithms/locate_point.hpp
int-ptr-ptr: I'll do both. The point of having an [edge|face]-local coordinate argument is to not require breaking into cases by `mesh_entity` each time we want to use the same coordinate system as `get_local_[edge|face]_coordinate`.

Rohit-Kakodkar (approved):

lsawade (approved):

icui (approved):

Issue #1863, opened 2026-05-18
Interpolation based coupling for NCI in 3D


lsawade:
## Tasks

- Write NCI kernel for interpolation-based coupling
- Add `chunk_face` analogues to `chunk_edge/nonconforming_interface` for kernel to work

## Acceptance Criteria

- None until tests

PR #1876, opened 2026-05-20
`...ChunkFaceViewType` types created + `...ChunkElementViewType` refactor

int-ptr-ptr:
## Description

- Creates `ChunkNDimViewType` to unify code.
- Creates `<Scalar|Vector|Tensor>ChunkFaceViewType` types similar to `chunk_edge`, delegating to `ChunkNDimViewType`.
- Refactors `chunk_element_view.hpp` to use `ChunkNDimViewType` (807 -> 179 lines)

## Issue Number

- defines base type to be used in #1863

## Checklist

### DIFF
--- a/core/specfem/datatype/chunk_face_view.hpp
+++ b/core/specfem/datatype/chunk_face_view.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "chunk_ndim_view.hpp"
+#include "specfem/datatype/accessor_type.hpp"
 #include <utility>
 
 namespace specfem {
@@ -13,6 +14,7 @@ template <typename T, specfem::element::dimension_tag DimensionTag,
 struct ScalarChunkFaceViewType
     : public ChunkNDimViewType<T, DimensionTag, NumberOfFaces,
                                NumberOfGLLPoints, 2, std::integer_sequence<int>,
+                               specfem::datatype::AccessorType::chunk_face,
                                UseSIMD, MemorySpace, MemoryTraits> {
   constexpr static auto accessor_type =
       specfem::datatype::AccessorType::chunk_face; ///< Accessor type for
@@ -28,8 +30,9 @@ template <
 struct VectorChunkFaceViewType
     : public ChunkNDimViewType<T, DimensionTag, NumberOfFaces,
                                NumberOfGLLPoints, 2,
-                               std::integer_sequence<int, Components>, UseSIMD,
-                               MemorySpace, MemoryTraits> {
+                               std::integer_sequence<int, Components>,
+                               specfem::datatype::AccessorType::chunk_face,
+                               UseSIMD, MemorySpace, MemoryTraits> {
   constexpr static auto accessor_type =
       specfem::datatype::AccessorType::chunk_face; ///< Accessor type for
                                                    ///< identifying the
@@ -46,8 +49,9 @@ template <typename T, specfem::element::dimension_tag DimensionTag,
 struct TensorChunkFaceViewType
     : public ChunkNDimViewType<
           T, DimensionTag, NumberOfFaces, NumberOfGLLPoints, 2,
-          std::integer_sequence<int, Components, NumberOfDimensions>, UseSIMD,
-          MemorySpace, MemoryTraits> {
+          std::integer_sequence<int, Components, NumberOfDimensions>,
+          specfem::datatype::AccessorType::chunk_face, UseSIMD, MemorySpace,
+          MemoryTraits> {
   constexpr static auto accessor_type =
       specfem::datatype::AccessorType::chunk_face; ///< Accessor type for
                                                    ///< identifying the
--- a/core/specfem/datatype/chunk_face_view.hpp
+++ b/core/specfem/datatype/chunk_face_view.hpp
@@ -1,0 +1,42 @@
+#pragma once
+
+#include "chunk_ndim_view.hpp"
+#include <utility>
+
+namespace specfem {
+namespace datatype {
+template <typename T, specfem::element::dimension_tag DimensionTag,
+          int NumberOfFaces, int NumberOfGLLPoints, bool UseSIMD = false,
+          typename MemorySpace =
+              Kokkos::DefaultExecutionSpace::scratch_memory_space,
+          typename MemoryTraits = Kokkos::MemoryTraits<Kokkos::Unmanaged>>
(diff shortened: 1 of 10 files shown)


int-ptr-ptr:
@Rohit-Kakodkar I wanted your thoughts on this.

core/specfem/datatype/chunk_face_view.hpp
```diff
@@ -0,0 +1,63 @@
...
+namespace specfem {
+namespace datatype {
+template <typename T, specfem::element::dimension_tag DimensionTag,
+          int NumberOfFaces, int NumberOfGLLPoints, bool UseSIMD = false,
+          typename MemorySpace =
+              Kokkos::DefaultExecutionSpace::scratch_memory_space,
+          typename MemoryTraits = Kokkos::MemoryTraits<Kokkos::Unmanaged>>
+struct ScalarChunkFaceViewType
+    : public ChunkNDimViewType<T, DimensionTag, NumberOfFaces,
+                               NumberOfGLLPoints, 2, std::integer_sequence<int>,
```
Rohit-Kakodkar, line 16:
```suggestion
                               NumberOfGLLPoints, specfem::element::dimension<DimensionTag>::dim, std::integer_sequence<int>,
```

core/specfem/datatype/chunk_ndim_view.hpp
```diff
@@ -0,0 +1,212 @@
...
+template <specfem::element::dimension_tag DimensionTag, bool UseSIMD>
+struct index;
+template <specfem::element::dimension_tag DimensionTag> struct face_index;
+template <specfem::element::dimension_tag DimensionTag> struct edge_index;
+} // namespace specfem::point
+
+namespace specfem::datatype {
+
+namespace impl {
+/**
```
Rohit-Kakodkar, line 21: Could this be simplified?

core/specfem/datatype/chunk_ndim_view.hpp
```diff
@@ -0,0 +1,212 @@
...
+      typename impl::arrayify_datatype<
+          typename specfem::datatype::simd<T, UseSIMD>::datatype, ChunkSize,
+          typename impl::subspace_shape<ChunkSubspaceDimension, NGLL>::type,
+          PointwiseShapeIntegerSequence>::type,
+      MemorySpace, MemoryTraits>; ///< Underlying data type used to
+                                  ///< store values
+  using value_type = typename type::value_type; ///< Value type used to store
+                                                ///< the elements of the array
+  using base_type = T;                          ///< Base type of the array
+  using index_type = std::conditional_t<
```
Rohit-Kakodkar, line 114: Should this section lie in another impl? index type is required for the derived chunk_element and chunk_face. So this could lie in another class and inherited within chunk_element and chunk_face - basically mixins.

core/specfem/datatype/chunk_ndim_view.hpp
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * @brief Get scalar value by a point index.
+   *
+   * @param index Point index
+   * @param icomp...
+   */
+
+  template <typename... ComponentTypes>
+  KOKKOS_INLINE_FUNCTION constexpr value_type &
+  operator()(index_type index, ComponentTypes... components) {
```
Rohit-Kakodkar, line 186: The operator could also be a mixin

Rohit-Kakodkar (changes requested):
This is great! 

Minor changes for clarity/maintenance.

core/specfem/datatype/chunk_face_view.hpp
int-ptr-ptr: Do you want face to be a 2D structure (in which this should stay 2), or do you want it to be a codimension-1 structure (1 in dim2, 2 in dim3)?

For the above code, I wrote it as the former, to which the magic 2 is canonical to `"face"`.

core/specfem/datatype/chunk_ndim_view.hpp
int-ptr-ptr: I'm not sure. The trouble is getting an integer sequence `<i1,i2,...,in>` to an array definition `T[i1][i2]...[in]`. I don't know of any fold expressions for array declarations, so it seems like it takes 2 passes.

The point of splitting into `SubspaceShape` and `PointwiseShape` is to allow the replacement of pointwise-indexing in the future:
```c++
  KOKKOS_INLINE_FUNCTION
  impl::VectorChunkElementSubview<VectorChunkElementViewType>
  operator()(const chunk_ndim_view_type::index_type &index) {
    return { *this, index };
  }
```
where `VectorChunkElementViewType` is derived from `typename PointwiseShapeIntegerSequence`.

core/specfem/datatype/chunk_ndim_view.hpp
int-ptr-ptr: This is a tough call. The point of passing in `specfem::datatype::AccessorType` is to set the `constexpr accessor_type` for container-loads, but `AccessorType` is sufficient to know which index type to use.

I just went with Congyue's strategy of forward-declaring them, so that indices do not get `#include`d.

The initial idea was to have `<Scalar|Vector|Tensor>Chunk<Edge|Face|Element>ViewType` just be defined as `using` aliases for the corresponding `ChunkNDimViewType`, since all of these data structures are the same:
- a `Kokkos::View<T[ixyz...][ijk...]>`
- inherited `operator()` for all-integer-accessing
- overloaded `operator()`s for index types
- various `constexpr` declarations used for template `load` and `store` operations

core/specfem/datatype/chunk_ndim_view.hpp
int-ptr-ptr: Since we expect every `<Scalar|Vector|Tensor>Chunk<Edge|Face|Element>ViewType` to have one of these, it made sense to include it into `ChunkNDimViewType`. Would it be easier to read if we instead had `ChunkNDimViewType` inherit an `indexing_mixin<DimensionTag,AccessorType,UseSIMD>` type?

lsawade (approved):

core/specfem/datatype/chunk_face_view.hpp
Rohit-Kakodkar: Okay. This is good.

Rohit-Kakodkar (approved):

icui (approved):

PR #1882, opened 2026-05-26
Remove GCC8 from jenkins.

icui:
GCC8 is deprecated and does not support some of the C++20 libraries.

### DIFF
--- a/.jenkins/gnu_compiler_checks.gvy
+++ b/.jenkins/gnu_compiler_checks.gvy
@@ -10,7 +10,7 @@ pipeline{
                 axes {
                     axis{
                         name 'GNUCompiler'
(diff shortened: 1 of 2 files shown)


.jenkins/gnu_compiler_checks.gvy
```diff
@@ -10,7 +10,7 @@ pipeline{
                 axes {
                     axis{
                         name 'GNUCompiler'
-                        values 'GCC8;gcc/11', 'GCC14;gcc-toolset/14'
```
Rohit-Kakodkar, line 13: GCC 8 seems to be a misnomer here. The module being loaded is GCC 11. Does this not compile?

PR #1883, opened 2026-05-27
Issue 1881 - Add parallel HDF5 support to VTK wavefield writer

lsawade:
## Description

- Add MPI-parallel HDF5 I/O to the dim3 VTK wavefield writer, supporting serial HDF5, serial HDF5 under MPI, and fully parallel collective HDF5 writes
- Update cmake/hdf5.cmake to auto-enable parallel HDF5 when MPI is on, detect parallel capability of system HDF5, and error/warn on MPI vs HDF5 mismatches
- Refactor the plotter to use long long for point/cell counts, store the assembly by reference, and extract reusable extend_and_write_scalar/extend_and_write_array helpers

### Supports 
- Build serial (no MPI, no HDF5) — compiles and runs
- Build serial + HDF5 — VTK HDF output is valid
- Build MPI + serial HDF5 — all ranks write to separate files, not ideal but works. Alternative is separate ranks write iteratively.
- Build MPI + parallel HDF5 — all ranks write collectively.

## Checklist


icui (approved):

core/specfem/periodic_tasks/plot_wavefield/dim3/plot_wavefield.cpp
```diff
@@ -26,6 +25,97 @@
...
+  h5_check(H5Sclose(dataspace), "H5Sclose", __LINE__, __FILE__);
+}
+
+} // namespace specfem::periodic_tasks::plot_wavefield_impl
+
+#define SPECFEM_H5_CHECK(call)                                                 \
+  specfem::periodic_tasks::plot_wavefield_impl::h5_check((call), #call,        \
+                                                         __LINE__, __FILE__)
+#define SPECFEM_H5_CHECK_ID(call)                                              \
+  specfem::periodic_tasks::plot_wavefield_impl::h5_check_id(                   \
```
icui: I don't think these macros should be in `plot_wavefield.cpp`

icui (changes requested):
I wonder if it if possible to generalize the code and move them to io/hdf5 because I feel a large chunk of the code will be written again once we need to implement parallel wavefield_writer.

core/specfem/periodic_tasks/plot_wavefield/dim3/plot_wavefield.cpp
lsawade: I think the only thing they would have in common is the generate wavefield on entire grid, no? Which would already be general. Most of the stuff that is happening here is mapping from our addressing to the VTK during the initialize phase, and then at each step the wavefield is mapped to a 1d array in a way that is readable by vtk with an offset. 
This is _very_ different from how we implemented our hdf5 wavefield writer in 2D.

I think the likely best way of implementing the same thing once is doing something similar to specfem3d_globe hdf5 and write a geometry xml file on the side. that way both adios and hdf5 could share identical structure too. There's just very little information about it.

lsawade: What I mean is if we do it that way, I think there may be a way to deprecate VTKHDF entirely and just use the hdf5 wavefield writer and write one extra file.

lsawade: I realize one more thing the wavefield writer and plotter are doing different things. the wavefield writer writes the literal wavefield whereas the plotter actual creates a writeable field derivative.

Rohit-Kakodkar (approved):

core/specfem/periodic_tasks/plot_wavefield/dim3/plot_wavefield.cpp
Rohit-Kakodkar: IMO, it makes sense that a writer and a plotter do different things. Since the goals are different, in writer we want the exact values at all GLL points. In the plotter, we want to make sure that the visualized wavefield is correct, which involves interpolating the values in the interior of the element before sending to OpenGL (or Mesa) for plotting.

Rohit-Kakodkar: To @icui point, maybe the (parallel HDF5) functions could be generalized better. The way they are rn they are only localized to the plotter.

lsawade:
@icui rerequested

icui (approved):

PR #1884, opened 2026-05-27
Add compile-time grid extents and tail-skip to chunked iterators

Rohit-Kakodkar:
## Description

This reduces some register usage on GPU by using compile time div-mods

- Add a `Grid<...>` tag and specialize `element_grid` on it so GLL extents can be baked in at compile time. The runtime `Grid<>` form still works; `Grid<N>` / `Grid<Nz,Nx>` / `Grid<Nz,Ny,Nx>` turn `ngllx`, `size`, etc. into `static constexpr`.
- Plumb the grid tag and a `ChunkSize` template parameter through the chunked and mapped-chunked iterators so the div/mod math that maps a policy index to (ielement, ix, iy, iz) uses literal divisors when the extents are known at compile time.
- Allow a fixed compile-time chunk size by giving chunk indices an `is_end()` sentinel and having `for_each_level` skip those positions, so partial trailing chunks are handled correctly. Edge, face, and intersection indices get a trivial `is_end() == false` for uniformity.
- In `compute_stiffness_interaction`, build the `element_grid` locally with the compile-time `Grid<NGLL>` extent and load `PointWeights` once up front instead of reloading it inside the Cosserat branch.
- `scattered_divergence` now takes the precomputed weight product rather than the weights view, dropping redundant `weights(...)` reads from the inner atomic-add loops.

## Issue Number

## Checklist

### DIFF
--- a/core/specfem/algorithms/scatter.hpp
+++ b/core/specfem/algorithms/scatter.hpp
@@ -41,14 +41,13 @@ namespace impl {
  * @param acc          Shared scratch acceleration view (non-const, atomic
  * writes)
  */
-template <typename TensorPointViewType, typename WeightsType,
-          typename QuadratureType, typename ChunkAccType>
+template <typename TensorPointViewType, typename T, typename QuadratureType,
+          typename ChunkAccType>
 KOKKOS_FORCEINLINE_FUNCTION void scattered_divergence(
     const TensorPointViewType &F,
     const specfem::point::index<specfem::element::dimension_tag::dim2,
                                 TensorPointViewType::using_simd> &local_index,
-    const WeightsType &weights, const QuadratureType &hprime,
-    ChunkAccType &acc) {
+    const T &w_product, const QuadratureType &hprime, ChunkAccType &acc) {
 
   constexpr int ngll = ChunkAccType::ngll;
   constexpr int components = TensorPointViewType::components;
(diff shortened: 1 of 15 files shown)


core/specfem/algorithms/scatter.hpp
```diff
@@ -88,14 +85,13 @@ KOKKOS_FORCEINLINE_FUNCTION void scattered_divergence(
...
-          typename QuadratureType, typename ChunkAccType>
+template <typename TensorPointViewType, typename T, typename QuadratureType,
+          typename ChunkAccType>
 KOKKOS_FORCEINLINE_FUNCTION void scattered_divergence(
     const TensorPointViewType &F,
     const specfem::point::index<specfem::element::dimension_tag::dim3,
                                 TensorPointViewType::using_simd> &local_index,
-    const WeightsType &weights, const QuadratureType &hprime,
-    ChunkAccType &acc) {
+    const T &w_product, const QuadratureType &hprime, ChunkAccType &acc) {
```
lsawade, line 94: Does this need to be any type because of simd?

lsawade (approved):
Awesome

core/specfem/algorithms/scatter.hpp
Rohit-Kakodkar: No you're right. It should be type_real

icui (approved):

PR #1885, opened 2026-05-27
Implements both kernels | choose correct on shared

Rohit-Kakodkar:
## Description

Restores the old kernel and choose correct kernel based on memory size

## Issue Number

## Checklist

### DIFF
--- a/core/specfem/compute/impl/compute_stiffness_interaction.hpp
+++ b/core/specfem/compute/impl/compute_stiffness_interaction.hpp
@@ -10,34 +10,6 @@
 #include <Kokkos_Core.hpp>
 #include <limits>
 
-// ---------------------------------------------------------------------------
-// Helper: query GPU max shared memory per block, cached per instantiation.
-// Returns std::numeric_limits<int>::max() on host builds so that the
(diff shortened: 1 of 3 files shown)


lsawade:
good thing that most of the internal machinery is being tested, otherwise I would not be for a two kernel implementation and rather swallow the cost

core/specfem/compute/impl/compute_stiffness_interaction.hpp
```diff
@@ -1,23 +1,47 @@
...
+    int v;
+    hipGetDevice(&device);
+    hipDeviceGetAttribute(&v, hipDeviceAttributeMaxSharedMemoryPerBlock,
+                          device);
+    return v;
+  }();
+  return value;
+#else
+  return std::numeric_limits<int>::max();
+#endif
```
icui: Why anonymous namespace? Also, should these functions be in `execution` or `parallel_configuration`?

icui (changes requested):

icui (approved):

lsawade (approved):

PR #1915, opened 2026-06-10
Issue 1869 - Adds a GMSH mesh cookbook

Rohit-Kakodkar:
## Description

- Adds the cookbook
- Adds a export gmsh script for 3d
- Fixes read receivers

image

image

image

@int-ptr-ptr Feel free to update this export script or remove it when you make an update for adding non-conforming interfaces using GMSH. 

## Issue Number

## Checklist


scripts/export_gmsh3d.py
```diff
@@ -0,0 +1,642 @@
+#!/usr/bin/env python
```
lsawade, line 1: Should this go in scripts?

lsawade (approved):

scripts/export_gmsh3d.py
Rohit-Kakodkar: Yes, this script exports any GMSH generated mesh to xdecompose

lsawade (approved):
I think you should make a parameter files folder with all files needed to run the cookbook, like we are doing for the other cookbooks and the just reference, e.g., `create_mesh.py` inside the .rst file. That way users can download all the necessary files at once.

lsawade (changes requested):
I think you should make a parameter files folder with all files needed to run the cookbook, like we are doing for the other cookbooks and the just reference, e.g., create_mesh.py inside the .rst file. That way users can download all the necessary files at once.

see marmousi example and attenuation examples. Note that not all cookbooks seem to be updated yet.

lsawade (approved):

icui (approved):

PR #1940, opened 2026-06-17
Fix stacey boundary condition configuration.

icui:
## Description

Fixes the bug the Stacey configuration does not respect `TOP_FREE_SURFACE` and `BOTTOM_FREE_SURFACE`.


Rohit-Kakodkar (approved):

lsawade (approved):
~This looks good to me, but did you not need to update the Mesh_parfiles to recreate the databases?~ Seems to be already there...

PR #1957, opened 2026-06-25
Issue 1810 - Implements Attenuation-aware IO

### LINKED ISSUE #1810: Fix attenuation properties I/O
## Task description

I have implemented the properties so that they are stored without attenuation parameters (Q). The

The properties I/O needs to be refactor to account for attenuation media.

* rescale the stored properties
* get Q for the GLL point in question
* store reference frequency in the file attributes so that the relaxation rates can be recomputed.

## Acceptance criteria

- [ ] Q is written to file.

lsawade:
## Description

Adds 
- AttenuationIO to `io/impl/`
- appropriate member functions in `assembly/attenuation`. To write Q and recompute the appropriate things upon reading
- adds `reader.read()` call to the assembly (both 2 and 3D were missing it)

## Issue Number

Closes #1810 

## Checklist


lsawade:
This one has severe merge conflicts. I'm working on it. I forgot to merge devel earlier.

core/specfem/io/impl/medium_writer.tpp
```diff
@@ -184,8 +194,16 @@ void specfem::io::impl::write_container(
...
+        // Per-view write hook: lets the property writer un-scale attenuating
+        // kappa/mu to physical moduli; defaults to a verbatim dataset write.
+        n_written += write_medium_group(
+            group, mesh, element_indices, data_container,
+            [&](const std::string &name, const auto &view) {
+              extra.template write_view<medium_tag, property_tag>(group, name,
+                                                                  view);
+            });
+        // Append any extra datasets (e.g. attenuation Qkappa/Qmu) to the group.
+        extra.template attenuation_append<medium_tag, property_tag>(group);
```
Rohit-Kakodkar, line 206: This should lie within property writer calling function ``property/writer.tpp``. When Congyue initially refactored this writer, it was intended to capture shared parts within properties and kernels - attenuation seems like property specific implementation and could be added to the calling function

Rohit-Kakodkar (commented):
How is this not breaking kernel writer

Rohit-Kakodkar (commented):
As mentioned earlier, medium writer was implemented to capture the shared parts of properties and kernel writers. The change you are implementing currently is seem to me are specific to properties (maybe I am wrong). The main problem you are trying to solve is having to scale kappa and mu before writing - If I think of that as an independent problem as if I was writing it from scratch, i.e. without the baggage of ``medium_writer`` I would 

- Open the HDF5 file
- Compute x and z
- Write x and z 
- scale kappa and mu 
- write kappa and mu
- write the other properties 
- write attenuation specific details
- close 

So I would implement that  as an attenuation specialization to property writer/reader first and then look at what code I can share with medium writer. The reason I say it is because the above version does not require the changes you made in ``constant_isotropic`` (apart from a function for computing scaling factors) and the changes you made to attenuation_io. It would simplify the code a lot.

PR #1978, opened 2026-07-06
Issue 1946 Part III: Implements wavefield checkpointing strategy and memory buffer

icui:
## Description

- Creates a strategy class to decide when to create a checkpoint (currently only fixed time step)
- Creates memory buffer for displacement field (currently only on device, `store_buffer_in_host` option could be added in the future)

## Issue Number

Adds to #1946 

## Checklist


core/specfem/solver/forward_displacement_buffer.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include "specfem/setup.hpp"
+#include "specfem/tag_dispatch.hpp"
+#include <Kokkos_Core.hpp>
+#include <stdexcept>
+
+namespace specfem {
+namespace solver {
+
+/**
+ * @brief In-memory ring buffer holding @p window_size forward displacement
```
Rohit-Kakodkar, line 15: Window size is a misleading term here. window_index would be better. Window size means ``[time_step_start, time_step_end)``

core/specfem/solver/checkpointing_strategy.hpp
Rohit-Kakodkar, line 1: Can ``CheckPointingStrategy`` be a parent of ``periodic_tasks``? If not, why?

lsawade:
I thought a little about this and maybe it makes sense to create a `Task`, `Policy`, `Scheduler` combination that covers it all. Then instead of having a vector of periodic_tasks we would have a scheduler, and should_run could take assembly as input, meaning we could even trigger tasks if idk, the field is exploding etc. From chat with claude:

```cpp
// Single-file example: tasks + swappable scheduling policies in a
// timestep-driven simulation.  C++17, no dependencies.
//
//   g++ -std=c++17 task_scheduler_example.cpp -o sched && ./sched

#include <cstdint>
#include <iostream>
#include <memory>
#include <random>
#include <string>
#include <vector>

using SimTime = std::uint64_t;

// ---------------------------------------------------------------------------
// 1. Task: knows WHAT to do. Nothing about timing.
// ---------------------------------------------------------------------------
class Task {
public:
    virtual ~Task() = default;
    virtual void initialize() {}
    virtual void run(SimTime now) = 0;
    virtual void finalize() {}
};

// ---------------------------------------------------------------------------
// 2. SchedulingPolicy: knows WHEN to run. Nothing about what the task does.
//    Polling style: the scheduler asks every timestep.
// ---------------------------------------------------------------------------
class SchedulingPolicy {
public:
    virtual ~SchedulingPolicy() = default;
    virtual bool should_run(SimTime now) = 0;
    virtual void on_executed(SimTime /*now*/) {}   // feedback hook for adaptive policies
    virtual bool done(SimTime /*now*/) const { return false; }  // true => remove task
};

// --- Concrete policies ------------------------------------------------------

// Classic fixed interval: your old "periodic" behavior, now just one policy.
class EveryN : public SchedulingPolicy {
    SimTime period_;
    SimTime last_ = 0;
    bool ran_once_ = false;
public:
    explicit EveryN(SimTime period) : period_(period) {}
    bool should_run(SimTime now) override {
        return !ran_once_ || now - last_ >= period_;
    }
    void on_executed(SimTime now) override { last_ = now; ran_once_ = true; }
};

// Erratic timing: interval re-drawn after every execution from a seeded RNG
// (seeded => runs stay reproducible, which simulations usually require).
class JitteredInterval : public SchedulingPolicy {
    std::mt19937 rng_;
    std::uniform_int_distribution<SimTime> dist_;
    SimTime next_;
public:
    JitteredInterval(SimTime min, SimTime max, unsigned seed)
        : rng_(seed), dist_(min, max), next_(dist_(rng_)) {}
    bool should_run(SimTime now) override { return now >= next_; }
    void on_executed(SimTime now) override { next_ = now + dist_(rng_); }
};

// Finite repetition: shows how a policy can end a task (=> finalize + removal).
class NTimesEvery : public SchedulingPolicy {
    EveryN inner_;
    int remaining_;
public:
    NTimesEvery(int times, SimTime period) : inner_(period), remaining_(times) {}
    bool should_run(SimTime now) override {
        return remaining_ > 0 && inner_.should_run(now);
    }
    void on_executed(SimTime now) override {
        inner_.on_executed(now);
        --remaining_;
    }
    bool done(SimTime) const override { return remaining_ <= 0; }
};

// ---------------------------------------------------------------------------
// 3. ScheduledTask: the registration. Binds one task to one policy and holds
//    per-registration state. Owned by the Scheduler; task and policy are
//    siblings here -- neither knows about the other.
// ---------------------------------------------------------------------------
struct ScheduledTask {
    std::unique_ptr<Task> task;
    std::unique_ptr<SchedulingPolicy> policy;
    bool initialized = false;
};

// ---------------------------------------------------------------------------
// 4. Scheduler: driven by the simulation loop, once per timestep.
//    Iteration order = registration order (deterministic tiebreak).
// ---------------------------------------------------------------------------
class Scheduler {
    std::vector<ScheduledTask> tasks_;
public:
    void add(std::unique_ptr<Task> task, std::unique_ptr<SchedulingPolicy> policy) {
        tasks_.push_back({std::move(task), std::move(policy), false});
    }

    void step(SimTime now) {
        for (auto& st : tasks_) {
            if (!st.policy->should_run(now)) continue;
            if (!st.initialized) { st.task->initialize(); st.initialized = true; }
            st.task->run(now);
            st.policy->on_executed(now);
        }
        // Remove (and finalize) tasks whose policy says they're finished.
        for (auto it = tasks_.begin(); it != tasks_.end();) {
            if (it->policy->done(now)) {
                if (it->initialized) it->task->finalize();
                it = tasks_.erase(it);
            } else {
                ++it;
            }
        }
    }

    void shutdown() {                     // finalize everything still registered
        for (auto& st : tasks_)
            if (st.initialized) st.task->finalize();
        tasks_.clear();
    }

    bool empty() const { return tasks_.empty(); }
};

// ---------------------------------------------------------------------------
// Example tasks -- note they are timing-agnostic and reusable.
// ---------------------------------------------------------------------------
class PrintTask : public Task {
    std::string name_;
public:
    explicit PrintTask(std::string name) : name_(std::move(name)) {}
    void initialize() override { std::cout << "[init] " << name_ << "\n"; }
    void run(SimTime now) override {
        std::cout << "  t=" << now << "  " << name_ << " runs\n";
    }
    void finalize() override { std::cout << "[fini] " << name_ << "\n"; }
};

// ---------------------------------------------------------------------------
// 5. main: the simulation loop drives the scheduler with the sim timestep.
// ---------------------------------------------------------------------------
int main() {
    Scheduler scheduler;

    // Same task type, three different timing strategies:
    scheduler.add(std::make_unique<PrintTask>("physics   (every 2)"),
                  std::make_unique<EveryN>(2));
    scheduler.add(std::make_unique<PrintTask>("sensor    (jitter 3-6)"),
                  std::make_unique<JitteredInterval>(3, 6, /*seed=*/42));
    scheduler.add(std::make_unique<PrintTask>("warmup    (3x every 1)"),
                  std::make_unique<NTimesEvery>(3, 1));

    for (SimTime t = 0; t < 20; ++t) {   // the simulation loop
        std::cout << "-- step " << t << " --\n";
        scheduler.step(t);
        // ... advance the rest of the simulation here ...
    }

    scheduler.shutdown();
}
```

lsawade (approved):

PR #1986, opened 2026-07-07
Issue 1810 - Attenuation property IO

### LINKED ISSUE #1810: Fix attenuation properties I/O

lsawade:
## Description

Here we go again!

Should address the original comments and is not modifying the `medium_writer`.

The properties IO tests have also been simplified compared to the original Attenuation IO implementation.

## Issue Number

Closes #1810 

## Checklist


core/specfem/io/impl/medium_writer.tpp
```diff
@@ -122,7 +130,8 @@ int specfem::io::impl::write_medium_group(
   const int n_elements = element_indices.size();
 
   using DomainView3d =
```
Rohit-Kakodkar, line 132: This is wrong. It should be derived from data containers.

core/specfem/io/impl/medium_writer.tpp
```diff
@@ -14,19 +14,18 @@
...
-    const ElementIndicesType &element_indices,
-    const DataContainerType &data_container) {
+    const ElementIndicesType &element_indices) {
 
   const int ngllz = mesh.element_grid.ngllz;
   const int ngllx = mesh.element_grid.ngllx;
 
   using DomainView =
-      typename DataContainerType::domain_view_type::host_mirror_type;
+      specfem::datatype::DomainView<specfem::element::dimension_tag::dim2,
```
Rohit-Kakodkar, line 27: Same here

lsawade:
Retest this please

core/specfem/io/property/writer.tpp
```diff
@@ -31,9 +42,236 @@ void property_writer<OutputLibrary>::write(specfem::assembly::assembly<specfem::
...
+
+        // The property container spans the whole (medium, property) group;
+        // this combination's elements are the contiguous sub-block
+        // [offset, offset + count) of its views.
+        const int offset = element_range.begin_index() -
+                           data_container.element_range.begin_index();
+        const int count = element_range.size();
+
+        impl::write_coordinates(group, assembly.mesh, element_range);
+
```
Rohit-Kakodkar, line 85: I think this is still confusing. I suggest attenuation return names of properties that are attenuated. i.e. "kappa" & "mu". Any transformations that happen to the view kappa & mu should happen in this file so its easy to follow. I suggest that attenuation could return scaling factors that can be used here.

Rohit-Kakodkar (changes requested):

core/specfem/io/property/writer.tpp
lsawade: The reason I set it up like this is that this one way of performing the scaling. For a different attenuation type that scaling may be different. So, if we make hardcode the scaling here, we would have to run scaling here every time.

I agree that this bit is confusing.

core/specfem/io/impl/medium_writer.tpp
lsawade: This is a miss on my end. Cannot tell you how many times I interrupted claude to please not edit the medium_writer. I should have checked at the end a last time...

Issue #2001, opened 2026-07-17
Fortran bindings for MESHFEM3D Globe


Rohit-Kakodkar:
### 5.1 Fortran library `libglobe_model`
Built from `meshfem3D_models.F90` + the `model_*.f90` catalog + the shared code
they need (`shared_parameters`, reference-model splines, ellipticity splines,
topo/bathy reader, attenuation setup). Two entry points, both `ISO_C_BINDING`:

```fortran
! called once at SPECFEM++ assembly setup
subroutine models_init(model_config, comm) bind(C)
  ! model_config = the MODEL_CONFIG block SPECFEM++ read from the database header
  !   (name string + resolved selection codes + model flags).
  ! set the shared_parameters / meshfem_models_par module state from model_config,
  ! then meshfem3D_models_broadcast():
  !   topo/bathy, 1D reference, 3D mantle, crust, attenuation, scattering
  ! NO Par_file parsing here — the mesher already resolved the model.
  ! uses the provided MPI communicator (SPECFEM++'s), or reads per-rank

! called per GLL point at setup
subroutine get_model_point(iregion, idoubling, rmin, rmax, elem_in_crust, &
                           x, y, z,                                        &
                           rho, vpv, vph, vsv, vsh, eta,                   &  ! TISO Love velocities
                           vp_iso, vs_iso,                                 &  ! Voigt isotropic average (CFL, iso media)
                           Qmu, Qkappa,                                    &
                           cij, Gc_prime, Gs_prime, is_anisotropic) bind(C)   ! full + azimuthal aniso
```

`get_model_point` is `get_model`'s inner loop body
([`src/meshfem3D/get_model.F90:104-247`](https://github.com/SPECFEM/specfem3d_globe/blob/9c312cb2c991b47484a7f302775f4f01ed9470f8/src/meshfem3D/get_model.F90#L104-L247))
with the array writes, `ispec,i,j,k` indexing, `MODEL_GLL`, and attenuation
storage removed; it returns values instead of storing them. It reuses the exact
same `meshfem3D_models_get1D_val / get3Dmntl_val / get3Dcrust_val / getatten_val`
calls, so results are bit-for-bit the mesher's.

**Output contract — three tiers of material + one rule:**
- **TISO** — the five Love velocities `(vpv,vph,vsv,vsh,eta)` + `rho`.
- **Isotropic Voigt average** — `(vp_iso, vs_iso)`, the anisotropic→isotropic
[...]

PR #2013, opened 2026-08-06
Expose the SPECFEM3D_GLOBE model catalog as a callable oracle

Rohit-Kakodkar:
## What this does

Exposes the SPECFEM3D_GLOBE model catalog as a callable oracle, so SPECFEM++ can
obtain the material properties that the globe mesher assigns — evaluated at
SPECFEM++'s own GLL points — without reimplementing the catalog.

Addresses #2001. Targets `issue-1994` (the meshfem3D_globe integration branch)
rather than `devel`, since it builds directly on the vendored globe tree added
there.

## Why an oracle rather than a port

The globe model catalog is ~90 `model_*.f90` files. Porting it to C++ is not
tractable, and any partial port would silently diverge from the mesher's own
`get_model` — producing plausible but wrong material rather than an error. This
PR calls the existing Fortran instead, and treats bit-for-bit agreement with the
mesher as the contract to defend.

## Layers

**`fortran/globe_oracle/`** — an ISO_C_BINDING wrapper over the vendored catalog.

- `globe_oracle_init` configures the catalog from a model **name** alone and lets
  `get_model_parameters()` re-derive every model flag and discontinuity radius.
  No globe `Par_file` is parsed anywhere in SPECFEM++, and there is exactly one
  place a model is chosen (the mesher, which writes the name into the database).
- `globe_oracle_get_element` evaluates a whole element in one call. This is
  required, not a convenience: the catalog's crustal `moho`/`sediment` state is
  element-scoped and genuinely carries across GLL points, so a point-by-point API
  would not reproduce the mesher for `ATTENUATION_3D` models.
- Kept in a sibling directory so `fortran/meshfem3d_globe/` stays a clean mirror
  of upstream specfem3d_globe.

**`core/specfem/globe_model/`** — `specfem::globe_model::Oracle`, a C++ facade.

- Confines the raw Fortran ABI to one translation unit; the public header is
  plain C++.
- Enforces the catalog's global-module-state constraint: the configuration lives
  in Fortran module variables, so a second live `Oracle` would share one model.
  Constructing one throws instead.
- Deliberately **not** part of `specfem::core` and not on the specfem
  executable's link line. It is the only target under `core/` that pulls in a
  Fortran archive and the Fortran runtime, so it stays opt-in until the assembly
  property build actually consumes it — which also keeps the coverage preset
  unaffected.

**`tests/unit-tests/models/`** — 21 tests over PREM.

Ground truth is the catalog's own PREM routines, via a test-only
`prem_reference` entry point, rather than a hand-written table of expected
values — a table would only confirm that it agrees with itself. Coverage:
configuration from a name (incl. case-insensitivity and the transversely
isotropic variant), the single-instance guard and post-destruction state,
rejection of models needing per-GLL indexing, malformed coordinate arrays,
full radial profiles against the reference, discontinuity handling across
adjacent shells, the outer-core acoustic tag, SI rescaling, direction
independence for 1D models, and the attenuation band.

PREM is analytic, so the suite consults no `DATA/` tree and runs from any
working directory.

## Notable design decisions

**Non-dimensional output.** Evaluated material is returned in the globe's own
non-dimensional units (density near 1, velocities near 2), not SI. Handing back
the catalog's numbers untouched is what keeps them bit-for-bit the mesher's;
converting inside the oracle would introduce a rounding step the mesher does not
have. Callers re-dimensionalize in one place, through `Oracle::scales()`.

**Exact float comparisons in the tests.** `.agents/rules/testing.md` asks for
`EXPECT_NEAR` with explicit tolerances, and that is the right default. It is
deliberately not followed here: the oracle's entire value proposition is
reproducing `get_model` exactly, and a tolerance would hide precisely the drift
these tests exist to catch. `EXPECT_DOUBLE_EQ` (4 ULP) is used for values that
must agree because they are the same computation; exact `EXPECT_EQ` appears once,
for the outer-core acoustic tag, where zero is a flag rather than a computed
quantity. The rationale is recorded in a header comment in `prem.cpp`.

## Build and gating

Everything is behind `SPECFEM_BUILD_MESHFEM3D_GLOBE`; with it off, nothing here
is configured. The test target is registered in `tests/unit-tests/mpi.cmake`
rather than `serial.cmake` because the globe Fortran tree has no serial MPI stub,
so the oracle only ever builds in MPI configurations. It runs on a single rank —
setup is collective, but this does not exercise multi-rank broadcasts.

## Known limitations, documented at the API boundary

- **Unrecognized model names abort the process.** The catalog terminates with a
  bare Fortran `stop` (`get_model_parameters.F90:978-982`), which cannot be
  converted to an exception. `GLOBE_ORACLE_UNSUPPORTED_MODEL` is returned only
  for models that parse but require per-GLL indexing (`MODEL_GLL`, CEM,
  `HETEROGEN_3D_MANTLE`, `ATTENUATION_GLL`, EMC). An inconsistent `idoubling`
  aborts the same way, from `get_model_check_idoubling`. Converting these to
  status codes is a follow-up.
- **Teardown is partial.** `globe_oracle_finalize` releases what the oracle owns
  and clears the init guard, but state allocated inside individual `model_*.f90`
  files cannot be freed — upstream exposes no hooks. Re-initializing is reliable
  for the analytic 1D reference models and may fail for 3D models that allocate.
  Production configures once per process.
- **The caller must evaluate at the mesher's reference coordinates** (spherical,
  Moho-stretched, before topography and ellipticity displace the points).
  Evaluating at final deformed positions samples the wrong depth near the
  surface. This cannot be checked inside the oracle.
- `min/max_attenuation_period` is the one input not derivable from the model
  name: the mesher computes it in `rcp_set_compute_parameters`, which the oracle
  deliberately does not call. The database has to carry it; the defaults in
  `ModelConfig` are placeholders.

## Testing

```bash
cmake --preset <mpi-preset> -D SPECFEM_BUILD_MESHFEM3D_GLOBE=ON
cmake --build build/<preset> --target globe_model_prem_tests
ctest --test-dir build/<preset> -R globe_model_prem_tests
```

### DIFF
--- a/core/specfem/globe_model/oracle.hpp
+++ b/core/specfem/globe_model/oracle.hpp
@@ -1,0 +1,262 @@
+#pragma once
+
+#include <cstddef>
+#include <string>
+#include <vector>
+
+namespace specfem {
+namespace globe_model {
+
+/**
+ * @brief Quadrature sizes the SPECFEM3D_GLOBE model catalog was compiled with.
+ *
+ * Check these against SPECFEM++'s own NGLL before trusting any evaluated
+ * material: a mismatch means the oracle samples a different quadrature rule
+ * than the caller, which produces plausible but wrong properties.
+ */
+struct Dims {
+  int ngllx = 0;
+  int nglly = 0;
+  int ngllz = 0;
+  int n_sls = 0;
+
+  /** @brief Number of GLL points per element, @f$ N_x N_y N_z @f$. */
+  std::size_t points_per_element() const {
+    return static_cast<std::size_t>(ngllx) * static_cast<std::size_t>(nglly) *
+           static_cast<std::size_t>(ngllz);
+  }
+};
+
+/**
+ * @brief Resolved model selection, replayed into the Fortran catalog.
+ *
+ * The model is identified by @ref model_name alone; the catalog re-derives
+ * every model flag and every discontinuity radius from it. SPECFEM++ therefore
+ * never parses the globe `Par_file`, and there is exactly one place a model is
+ * chosen (the mesher, which writes this name into the database).
+ */
+struct ModelConfig {
+  std::string model_name;
+
+  /** @brief Path for the catalog's own log output. Empty sends it to
+   * /dev/null. */
+  std::string log_path;
+
+  int planet_type = 1; ///< IPLANET_EARTH
+  int nchunks = 6;
+  int nex_xi = 64;
+  int nex_eta = 64;
+
+  bool ellipticity = false;
+  bool topography = false;
+  bool oceans = false;
+  bool attenuation = false;
+  bool gravity = false;
+  bool rotation = false;
+
+  /// @name Attenuation period band, in seconds
+  /// @{
+  /// Consulted only when @ref attenuation is set. Unlike everything else here,
+  /// this is *not* derivable from @ref model_name: the mesher computes it in
+  /// `rcp_set_compute_parameters`, which the oracle deliberately does not call,
+  /// so the database has to carry it. The defaults below are typical global
+  /// values and a placeholder -- production must take them from the mesher.
+  double min_attenuation_period = 20.0;
+  double max_attenuation_period = 1000.0;
+  /// @}
+};
+
+/**
+ * @brief Factors converting the oracle's output to SI.
+ *
+ * The catalog works in -- and @ref ElementProperties therefore reports -- the
+ * globe's non-dimensional units, where density is near 1 and velocities near 2.
+ * That is deliberate: handing back the catalog's own numbers untouched is what
+ * makes them bit-for-bit the mesher's, whereas converting inside the oracle
+ * would add a rounding step the mesher does not have. Callers re-dimensionalize
+ * here, in one place.
+ *
+ * \f[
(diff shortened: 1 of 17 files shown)


icui:
How about renaming `Oracle` to `ExternalModel`? I feel oracle is a bit too fancy and could be confused with the company name.

core/specfem/globe_model/oracle.hpp
```diff
@@ -0,0 +1,262 @@
...
+  /// @{
+  /// Consulted only when @ref attenuation is set. Unlike everything else here,
+  /// this is *not* derivable from @ref model_name: the mesher computes it in
+  /// `rcp_set_compute_parameters`, which the oracle deliberately does not call,
+  /// so the database has to carry it. The defaults below are typical global
+  /// values and a placeholder -- production must take them from the mesher.
+  double min_attenuation_period = 20.0;
+  double max_attenuation_period = 1000.0;
+  /// @}
+};
```
icui, line 67: `ModelConfig` needs to be dynamic eventually, these values have been saved in the thin mesh layer and create a `ModelConfig` instance during the reading phase.

icui (approved):

lsawade (approved):
nice!

PR #2025, opened 2026-08-12
Issue 1984 - Phase 3: Static-solver steady-state test and fixture suite

### LINKED ISSUE #1984: Implicit Solver implementation
- **`implicit_solver.hpp/.cpp`** — assemble `M` (lumped diagonal from
  `initialize_mass_matrix`) and later `C` (Stacey, probing the velocity path); form the
  implicit Newmark operator `A = M/(βΔt²) + (γ/(βΔt))C + K`; solve with **Belos GMRES**
  (non-symmetric once `C`/coupling present) + Ifpack2/MueLu preconditioner.
- Integration test: static elastic solve vs analytic/refined reference; compare to an
  explicit run driven to steady state.

lsawade:
> [!NOTE]
> **Stacked PR (3/3)** for the implicit Newmark solver (issue #1984), based on #2024 (`issue-1984-phase-2`). Only the last commit is new here. Merge order: #2002 → #2023 → #2024 → this.

## Description

The ImplicitNewmark integration suite gets its own data tree mirroring `Newmark/serial/dim3`: each fixture carries the full provenance (fortran DATA + specfempp mesh files) and a database-only Snakefile that regenerates `database.bin` via the shared rule modules (cross-suite include of `Newmark/specfem3d.smk`; hoisting the modules is a follow-up). These are not trace-regression fixtures — no `traces/` and no `tests.yaml` entry; the reference is always a live explicit run.

**Steady-state test** (the static solver of issue #1984): a Heaviside step force (hdur 4 s — large enough that the implicit dt = 0.7 does not trigger the Heaviside hdur ≥ 5 dt clamp, which would silently give the two runs different forcings) on the Stacey halfspace. The implicit solver taking 30 dissipative steps (γ = 0.6) recreates the explicit solver taking 600 steps to the same final time T = 21 s within 5% relative L2 on the full displacement and velocity fields. Fields are compared at the same T because the box converges to a constant-velocity drift, not a displacement fixed point. The detector assertion is mechanism-only (fires at a coarse tolerance, reusing the solver instance and its assembled operators via the new `set_steady_state_tolerance()`); where a tight tolerance fires depends on the fixture's physical ring-down, which the solver now logs per step when the detector is armed.

**Solver:** accept a non-Converged Belos return when the true residual b − A·u meets the tolerance — single-precision PseudoBlockGmres aborts on its internal loss-of-accuracy check while the solution is fine (observed: flagged at step 29 with true relative residual 2e-6 against a 1e-5 request).

**Also:** Heaviside branch in the dim3 YAML source parser, dynamic-equivalence fixture copied into the suite tree, unit tests merged into one process (construction dominates), and the integration tests labeled `long` (exclude locally with `ctest -LE long`).

## Issue Number

Closes #1984

## Checklist


lsawade:
These tests are expensive. I think we have to think about a different way of integration testing these.

Rohit-Kakodkar (commented):
What is this test? It seems like you are doing newmark integration converged to steady state with an assembled matrix. Does this also implement a steady state equation to solve? If its just doing a newmark integration then why do we need to use full matrix? What are we trying to achieve here - steady state or newmark or both?

Issue #1982, opened 2026-07-07
Assemble the stiffness matrix as a ``Tpetra::CrsMatrix<float>``


Rohit-Kakodkar:
New files under `core/specfem/linear_system/` (namespace `specfem::linear_system`).

1. **`dof_map.hpp`** — build the `Tpetra::Map` of owned global ids and the `gid(iglob,icomp)`
   helper from `get_iglob`/`get_nglob`. Owned iglob → owned GIDs; shared-interface points →
   column/overlap map; `fillComplete`'s `Export(ADD)` reproduces the matrix-free assembly sum.
2. **`element_stiffness.hpp/.tpp`** — per element, form the small dense block `K_e`
   (size `(NGLL³·ncomp)²`) by probing the single-element operator with local unit vectors,
   reusing `element_gradient`/`compute_stress`/`element_divergence`, velocity ≡ 0 (pure `K`),
   no mass division. Scalar = `float`.
3. **`tpetra_assembler.hpp/.cpp`** — build `Tpetra::CrsGraph` from element connectivity,
   then fill `Tpetra::CrsMatrix<float>` directly by scattering each `K_e` (batched row
   inserts; only the small `K_e` block is ever materialized — no global dense matrix),
   `fillComplete()`.

Scope guards: 3D single-medium **elastic isotropic**, `attenuation_tag::none`,
`boundary_tag::none`/free-surface only. Defer Stacey (`C`) and fluid-solid coupling.

PR #2078, opened 2026-09-02
Sparse matrix view: mesh-coordinate matrix updates

Rohit-Kakodkar:
Wraps the Tpetra matrix and vector fills behind a view indexed in mesh
coordinates, converts all six value-update sites onto it, and removes the dof
numbering stack that the migration leaves unreachable.

Built on top of #1982's `issue-1982` branch; the first two commits
(`6c328dfc5`, `861faf6d1`) are the branch's existing work.

## The problem

Filling a matrix was a raw-pointer, row-at-a-time operation repeated at six
sites, each re-implementing the same invariant. The stiffness fill was
representative: a `std::vector` of column ids allocated per element, a loop
decomposing one dense element block into `ndof_e` independent row calls, a raw
`&h_k_e(e, r, 0)` whose correctness depended on a layout declared twenty lines
away, and eight lines of hand-composed error handling because
`sumIntoGlobalValues` silently drops entries outside the graph and merely
returns a count.

## What it looks like now

```cpp
// stiffness: one block-diagonal update per probe batch
const auto dofs = mapping(batch, ALL, ALL, ALL, ALL);
K(dofs, dofs) += h_k_e;

// the implicit Newmark operator, A = K + c1 C + c2 M
A += *stiffness_;
A += damping_coefficient * (*damping_);
A += mass_coefficient * diag(*mass_);
```

`form_operator` goes from 65 lines of three row-copy passes to three
statements. `M` is a diagonal matrix that SPECFEM stores as a vector, so
`diag()` — MATLAB's spelling — says what the operator is rather than exposing
the storage through a `diagonal()` accessor.

## Pieces

- **`DofSet`** (Trilinos-free). Every index slot takes an integral,
  `Kokkos::ALL`, or a container of indices, so one selection names one dof, one
  element, or a set of either. Nothing is expanded at construction.
- **`SparseMatrixView` / `BlockRef`.** Owns the graph invariant, the column
  expansion (one buffer per assembly, not one `std::vector` per element), and
  the block layout, which is a template parameter rather than an assumption
  about a pointer. Block rank disambiguates a multi-element selection: rank 2
  is one dense block, rank 3 is one diagonal block per leading index — already
  the shape the stiffness probe produces, so the per-element scatter loop
  disappears rather than being rewritten.
- **`copy_field_to_vector` / `copy_vector_to_field`.** `Mapping` numbers dofs so
  that `gid(iglob, icomp)` is exactly the `LayoutLeft` offset of
  `(iglob, icomp)`, which is how a field is stored — so a field and a solver
  vector are the same bytes in the same order. The four hand-written double
  loops that copied them across one element at a time (and strided by `nglob`
  on *both* sides, twice per timestep) become one `deep_copy`, with an indexed
  fallback when the layout contract cannot be confirmed.
- **Deprecation.** `dof_map.hpp`, `dof_numbering.hpp`, `tpetra_connections.hpp`
  and `trilinos_smoke` are deleted — 700 lines with no reachable caller. The
  three-way split existed so the numbering half named no solver type; that
  property survives, since `mapping.hpp` and `dof_set.hpp` carry no Trilinos
  guard while every Teuchos/Tpetra type lives in `FEAssembly`.

The solver now builds one `FEAssembly` and threads it into both assemblers
instead of each building its own, halving setup-time graph construction from
four graphs to two.

## Two tests had stopped testing anything

`FullGraphMatchesStiffnessAssembler` and `DampingGraphMatchesDampingAssembler`
cross-checked `FEAssembly`'s graphs against the old `build_graph` path. Once
both sides were built by `FEAssembly` they compared a graph to an
identically-constructed one, and the damping test's comment still claimed the
two masks were derived independently. The first is replaced by a structural
clique-per-element assertion; the second is deleted, as
`DampingGraphIsBlockDiagonal` and `DampingGraphAgreesWithMask` already cover
it. Where the damping assembler now takes its sparsity from boundary tags
rather than the probe's nonzero pattern, that agreement is enforced at runtime
by an explicit throw instead.

## Verification status — please read before merging

**Nothing here has been compiled against Trilinos, and no test in this module
has been run.** Two independent reasons:

- No Trilinos install was available on the development machine, so everything
  behind `SPECFEM_ENABLE_TRILINOS` is unbuilt. The `FECrsMatrix` API this
  assumes — construction from `RCP<const FECrsGraph>`, `beginAssembly`,
  `endAssembly` — needs confirming. If it differs, the fallback is a plain
  `crs_matrix_type` on the graph plus `fillComplete`, localized to
  `SparseMatrixView`'s constructor, `begin_fill` and `finalize`.
- The 3D displacement-test datasets are absent from the checkout, so every test
  in this module fails on a missing `specfem_config.yaml` before reaching an
  assertion. This is pre-existing and unrelated, but it means the new tests
  have not executed.

What *was* verified: the full non-Trilinos tree builds clean; all nine test
targets link; and the two load-bearing contracts were checked at runtime
against stub mappings — that `DofSet`'s ordering matches `local_dof_index`
(with `icomp` slowest, so the written slot order deliberately is not C-order),
and that the flat field/vector copy is identical to the indexed gather.

The decisive existing tests are untouched and are what should be trusted on
first build: `K·(rigid translation) == 0` and `K·u` against the matrix-free
kernel, the damping matrix against the production kernel, and
`SystemOperatorMatchesKCM`, which computes `K·x + c1 C·x + c2 M∘x`
independently.


Rohit-Kakodkar:
Superseded by a three-PR stack, which should be much easier to review:

- #2079 — **1/3 interface**: `DofSet`, `SparseMatrixView`, the field/vector copies, and the six call sites converted onto them. Production code only, plus the five-line `implicit_solver_tests` change forced by the `dof_map()` → `fe()` rename.
- #2080 — **2/3 tests**: the two new test files and their CMake registrations.
- #2081 — **3/3 deprecation**: deletes the now-unreachable dof numbering stack and `trilinos_smoke`, and replaces two tests that had become tautological.

Each builds the full tree on its own, and the tip of the stack reproduces this branch's tree byte-for-byte. Closing in favour of those.

Issue #2066, opened 2026-08-24
Use TensorOps for StaticSolver stiffness kernel


lsawade:
## Task description

Test the TensorOps 

## Acceptance criteria
- [ ] TensorOps computes the StiffnessKernel

PR #2092, opened 2026-09-17
Issue 2066 - Assemble element stiffness through a declarative TensorOperations graph (#2066)

lsawade:
Closes #2066.

Adds a second, selectable element-stiffness kernel to
`specfem::linear_system`: a declarative
[TensorOperations](https://github.com/Rohit-Kakodkar/TensorOperations) level
graph producing the dense `K_e` blocks for the implicit solver. The
unit-vector **probe** kernel stays as the always-available correctness
oracle; builds without the dependency are bit-identical to before.

### Design

- `StiffnessKernelImpl { probe, tensor_graph }` selects the `K_e` producer;
  `tensor_graph` is the default under `SPECFEM_ENABLE_TENSOROPS=ON`, `probe`
  otherwise.
- The kernel is one LevelGraph mirroring the weak form — gradient
  contractions → pointwise constitutive combine → weighted divergence — and
  assembly is the action evaluated on the identity via an extra blocked
  column label, so the graph output *is* `K_e`. The constitutive combine
  delegates to `algorithms::chain_rule` and the tag-dispatched
  `medium_physics::compute_stress`; it is the only medium-specific node.
- `StiffnessTensorGraphKernel` owns a persistent device workspace, allocated
  once per assembly at batch capacity; `make_element_stiffness_kernel<Tags>`
  binds a kernel once, keeping the assembler's fill loop kernel-agnostic. Its
  `k_e` device view is the documented seam for the device-scatter follow-up.
- `chain_rule` and `transverse_weighted_sum` are extracted into
  `specfem::algorithms`, with the loop-form algorithms and the graph functors
  both delegating. The dispatcher is preprocessor-free: OFF builds get
  throwing stubs, and exactly one TU includes TensorOperations headers.
- The graph deliberately deviates from the `specfem::execution` iterator
  idiom — TensorOperations owns its own team scheduling, and piloting it on a
  production path is the point of #2066 (recorded in the kernel doxygen).

### Dependency

TensorOperations is header-only and never vendored or fetched (no license
yet; its own CMake pulls a second Kokkos). A checkout is supplied via
`SPECFEM_TENSOROPS_ROOT`; `cmake/tensorops.cmake` builds an interface target
over it against SPECFEM++'s Kokkos.

### Testing

- Fixture-free smoke test pinning the exact library features the kernel
  needs — a Kokkos or TensorOperations bump breaks here first, in isolation.
- A/B test: `K_graph == K_probe` over all 288 fixture elements at
  scale-relative tolerance, plus symmetry and rigid-body null-space checks.
- Existing element/assembler `K·u` oracles and the `ImplicitNewmark3D`
  end-to-end tests run on the tensor-graph default and stay green; OFF builds
  compile with bit-identical probe behavior (a `tensor_graph` request
  throws).

### Performance

Host-serial: probe ~450 ms vs graph ~870 ms over 288 elements — the host
column tile is scratch-capped at 3; the GPU tile of 15 is the target, and a
GPU measurement is the next campaign. The assembler no longer pays per-batch
workspace allocation.

### Follow-ups (issues to be filed)

Device-side CRS scatter of `k_e` (consumer-side only — the kernel contract is
already the seam); TensorOperations broadcast labels (upstream; removes the
materialized identity); medium-agnostic graph (deferred until a second
medium); linear_system architecture page; shared test-support for the suite;
upstream Debug-build `index_consistent()` fix.


core/specfem/linear_system/impl/stiffness_tensor_graph_kernel.cpp
```diff
@@ -0,0 +1,441 @@
...
+  // The two 1D derivative operators the contractions stage: the gradient's
+  // hprime(point, function) and the divergence's transposed
+  // hprime(summed, point) with the summed point's quadrature weight folded in
+  // (element_divergence keeps weights(l) inside the sum).
+  Kokkos::parallel_for(
+      "specfem::linear_system::tensor_graph::stage_operators",
+      Kokkos::MDRangePolicy<ExecSpace, Kokkos::Rank<2>>({ 0, 0 },
+                                                        { NGLL, NGLL }),
+      KOKKOS_LAMBDA(const int p, const int f) {
+        derivative(p, f) = hprime(p, f);
```
Rohit-Kakodkar, line 173: Why do you need to prepopulate these?

Rohit-Kakodkar (commented):
I thought the point of using the tensor operations library was to use closed form expression and not having to probe the explicit implementaiton with unit vectors.

lsawade:
> I thought the point of using the tensor operations library was to use closed form expression and not having to probe the explicit implementaiton with unit vectors.

@Rohit-Kakodkar I tried this, but the results is that I need to keep a 5d array alive during the population of the matrix. I tried this earlier, based on your prior research, and the resulting code became _very_ ugly. Let me see if I can find it

PR #2095, opened 2026-09-21
VectorView: MATLAB-style vector expressions for the implicit solver

Rohit-Kakodkar:
Follow-up to the sparse-matrix-view stack (#2079, #2080, #2081). Those gave
the matrix side a mesh-coordinate grammar; this does the same for vectors.

## The problem

`ImplicitNewmarkSolver::run()` implements the displacement-form Newmark
update, but the algebra was unreadable at the call site: every equation was a
sequence of `update` calls threaded through two scratch members, so you
reconstructed the math by tracking `tmp_` and `tmp2_` across statements.

```cpp
tmp_->update(c_c0, *u_, c_c1, *v_, 0);
tmp_->update(c_c2, *a_, 1);
damping().apply(*tmp_, *tmp2_);
rhs_->update(1, *tmp2_, 1);
```

## After

```cpp
b += diag(mass()) * (c_a0 * u + c_a1 * v + c_a2 * a);
b += damping() * (c_c0 * u + c_c1 * v + c_c2 * a);

a_new = c_a0 * (u_new - u - dt * v) - c_a2 * a;
v_new = v + c_v0 * a + c_v1 * a_new;

const type_real residual_norm = norm2(b - system_operator() * u_new);
```

14 arithmetic lines become 4. `tmp_`/`tmp2_` are gone, replaced by the
space's scratch pool holding the same two vectors, so allocation is unchanged.

## Pieces

- **`VectorSpace`** owns a dof map and a scratch pool. Vectors come from a
  space rather than a map so an expression can borrow a temporary without
  allocating one per statement. This is also what lets `norm2(b - A*x)` --
  which has no assignment target -- find scratch.
- **`VectorView`** is an owning vector whose assignment takes an expression.
  Borrow-only proxies (`Sum`, `MatrixProduct`, `DiagProduct`) are flattened
  into Tpetra calls at the assignment; no temporary vector per operator.
- **`operands.hpp`** extracts `ScaledMatrix` / `Diagonal` / `ScaledDiagonal` /
  `diag()` out of `matrix_view.hpp` verbatim so both views share them. No API
  change -- `matrix_view.hpp` includes it.

## Grammar, deliberately minimal

A scaled sum of any length plus **at most one** matrix or diagonal product --
exactly what `run()` needs. Everything else is a `static_assert` naming the
fix rather than a silent allocation:

```cpp
b += 2.0f * C * u;      // "Scale the operand instead: A * (alpha * x)"
b = A * (B * u);        // "nested matrix products are not supported"
b = A * u + B * v;      // "an expression carries at most one matrix product"
```

Scaled operators, nested products, `dot()` and elementwise division are out
until a caller needs them. `SparseMatrixView::operator+=` keeps accepting
`ScaledMatrix`/`ScaledDiagonal` exactly as today -- that path is untouched.

## Emitted calls

Terms are emitted two at a time into `update`, and `apply` / `elementWiseMultiply`
accumulate through their own `beta`/`gamma`, so the sequence is never longer
than the hand-written one:

| statement | now | before |
|---|---|---|
| `b += diag(M) * (3 terms)` | 3 | 3 |
| `b += C * (3 terms)` | **3** | 4 |
| `a_new = c0*(u_new-u-dt*v) - c2*a` | **2** | 3 |
| `v_new = v + c0*a + c1*a_new` | 2 | 2 |
| `norm2(b - A*u_new)` | 3 | 3 |

An expression that reads its own target (`x = 2*x + y`) is routed through
scratch rather than being order-dependent.

## Verification status -- please read

**Not compiled against Trilinos**: no install was available on the development
machine, so everything behind `SPECFEM_ENABLE_TRILINOS` is unbuilt. What
still needs confirming on a cluster build is that Tpetra's `update`,
`elementWiseMultiply`, and `apply(X, Y, mode, alpha, beta)` behave as assumed
-- specifically that the latter two accumulate through `beta`/`gamma`, which
is what makes the call counts above come out.

What was verified: the full non-Trilinos tree builds clean, and the grammar
was exercised against a stub Tpetra that records every call --
20 assertions on emitted call *sequences* and values (sums of 1-5 terms,
assign vs accumulate, both products, `norm2`, aliasing, swap, pool growth),
plus a second harness replaying `run()`'s statements verbatim and checking
them entry-by-entry against the hand-evaluated Newmark equations. All four
rejected forms were confirmed to produce their intended `static_assert`.

The decisive existing tests are untouched and are what should be trusted on
first cluster build: `ConstructsAndOperatorMatchesOnStaceyMesh` and the
Newmark displacement regression. The new `vector_view_expression_tests` are
mesh-free -- they build a map directly rather than assembling a simulation --
so they run in milliseconds.


Rohit-Kakodkar:
Ran a cleanup pass over the three commits (reuse / simplification / efficiency / altitude). Pushed as `4785b599a`. Net **-13 lines** in the grammar headers.

**One correction to the PR description above.** The "never worse" claim was false. `norm2(b - A*u_new)` desugars with an inner `alpha` of `-1`, which missed the `alpha == 1` fast-path guard, so it scaled into scratch first: **5 Tpetra calls and 2 pool slots** against the original's 4 and 1. `apply` carries its own `alpha`, so the guard is gone and it is now 4 and 1. The corrected table:

| statement | now | before |
|---|---|---|
| `b += diag(M) * (3 terms)` | 3 | 3 |
| `b += C * (3 terms)` | **3** | 4 |
| `a_new = c0*(u_new-u-dt*v) - c2*a` | **2** | 3 |
| `v_new = v + c0*a + c1*a_new` | 2 | 2 |
| residual guard | 4 | 4 |

The scratch pool now never exceeds **one** vector — the description says it holds the same two as `tmp_`/`tmp2_`, so the new code is one Tpetra vector lighter in steady state, not equal.

**Also fixed**

- `b += A * b` reached Tpetra's `apply()` with input aliasing output, which is not permitted — the alias check ran only on the assigning path. Now routed through scratch on both, with a test.
- Three binary `operator-` collapse to one `left + (-right)`; `A*u - B*v` now gets the authored one-product message instead of a bare no-match.
- Scratch pool is a depth counter, not an occupancy vector plus scan (borrows are strictly LIFO).
- Three concepts (`VectorExpression`, `SumLike`, `VectorProduct`) replace five copies of the same `requires` clause and remove every `_impl` trait from a public signature.
- Rejected forms return a grammar type rather than `void`, so each is **one** diagnostic instead of two or three.
- Copy constructor deleted — assignment copies values, so a handle-sharing copy constructor made `VectorView b = u;` and `b = u;` mean opposite things.
- Solver's seven state vectors are `std::optional`, not `unique_ptr`.

**Deliberately not done — worth a reviewer's opinion.** The altitude review argued the one-product cap is arbitrary, and that a general term list would be *less* code than `Expression<N, Product>` plus its four operators and the reject. It also notes `Expression` is closed under nothing, so `2.0f * (b - A*u)` already fails with no guidance. I left it as-is because the minimal grammar was a deliberate call when this was planned, and widening it is a redesign rather than a cleanup — but it is the one finding I would not dismiss, especially with fluid-solid coupling coming.

Verification is unchanged from the description: still not compiled against Trilinos, still exercised against a stub Tpetra plus the non-Trilinos build.

lsawade (approved):

icui (approved):

tests/unit-tests/linear_system/vector_view/expression_tests.cpp
```diff
@@ -0,0 +1,395 @@
...
+    if (comm->getSize() > 1) {
+      GTEST_SKIP() << "the expression tests are single-rank.";
+    }
+    map_ = Teuchos::rcp(
+        new map_type(static_cast<Tpetra::global_size_t>(num_dofs), 0, comm));
+    space_ = std::make_unique<VectorSpace>(map_);
+  }
+
+  /// `alpha * I`, so that `A * x` has an obvious closed form
+  Teuchos::RCP<crs_matrix_type> scaled_identity(const scalar_type alpha) {
```
icui: Would it be better to return `VectorView` here?

icui (approved):
Minor comment.

PR #2096, opened 2026-09-21
Generalize the vector grammar to a term list

Rohit-Kakodkar:
Stacked on #2095. That PR's grammar carries **at most one** matrix or diagonal
product, through a dedicated `Expression<N, Product>`. This removes the cap.

## The problem isn't the cap

It's that `Expression` is closed under nothing. No operator accepts it, so the
moment an expression contains a product it cannot be composed further — and the
natural next steps fail with a *bare no-match*, no guidance:

```cpp
2.0f * (f - A*u)      // bare no-match
-(f - A*u)            // bare no-match
(f - A*u) + f         // bare no-match
norm2(f - A*u + f)    // bare no-match -- an ordinary residual with a source term
```

The grammar reads like algebra right up until it silently doesn't.

## After

`Sum` and `Expression` merge into one type — a term array plus a product tuple,
where a pure sum is zero products and a bare product is zero terms:

```cpp
template <std::size_t N, typename... Products> struct Expression {
  std::array<VectorTerm, N> terms;
  std::tuple<Products...> products;
};
```

Everything normalizes through `as_expression()`, after which **one `operator+`
replaces four** (sum+sum, sum+product, product+sum, and the two-product reject).
`emit()` pairs plain terms into `update()` exactly as before, then folds over
the products.

All four forms above now work, as do two products in one expression, mixed
product kinds, products over multi-term operands, and nested products.

## The rejected-forms section is deleted entirely

Nothing algebraic is left to turn away. Non-algebraic spellings — `A*B`,
`diag(m)*A`, `u + 1.0f`, `u*v`, `A + u`, `norm2(A)` — are excluded by the
operand concept alone, verified. That removes four `static_assert` overloads
and the `always_false_v` helper.

`ScaledMatrix`/`ScaledDiagonal` gain real operators instead of rejections:
`alpha*A*x` folds the coefficient into the operand, where it reaches the `alpha`
`apply()` already takes — one call, zero scratch. That is exactly what the old
error message told the caller to do by hand.

## Measured, not assumed

- **`run()` is untouched** — not one character, and its emitted Tpetra sequence
  is byte-identical. The diff touches two headers and the test file; nothing else.
- **All 13 existing tests pass unmodified**, which is the evidence this is a pure
  superset rather than a rewrite.
- Grammar operators **13 → 9**; `static_assert` rejections **5 → 0**.
- Compile time **537 → 450 ms** on the heaviest TU, best of 3.
- Scratch is bounded by product *nesting depth*, not by use: **1** vector for
  products side by side, **2** for a nested product, and repetition never grows
  the pool.

**Line count is flat, not smaller.** A scratchpad prototype suggested 467 → 323,
but it used short unqualified names; the project bans `using namespace` at
namespace scope, and the mutually recursive helpers need forward declarations.
Those account for the difference. The structural simplification is real; the
line saving was not, and I'd rather say so than quote the prototype number.

## Verification status

**Not compiled against Trilinos** — no install on this machine, same caveat as
#2095. Verified with the stub-Tpetra harnesses (which assert exact emitted call
sequences) and a clean non-Trilinos build. The decisive tests remain
`ConstructsAndOperatorMatchesOnStaceyMesh` and the Newmark displacement
regression on a cluster build.

## One behaviour change worth noting

`3 * A * u` changes from a compile error to working code. Strictly better, but
the "scale the operand instead" guidance disappears — anyone who learned that
rule learned one that no longer applies.


lsawade (approved):
I understand the puprose of this PR, but the amount of templating here makes it very difficult to follow.

tests/unit-tests/linear_system/vector_view/expression_tests.cpp
```diff
@@ -372,16 +372,168 @@ TEST_F(VectorExpression, ScratchIsPooledAcrossStatements) {
...
+  const auto b_matrix = scaled_identity(static_cast<scalar_type>(3));
+
+  fill(u.vector(), [](int i) { return i + 1; });
+  fill(v.vector(), [](int i) { return 10 * (i + 1); });
+
+  // Two products side by side still need only one scratch vector at a time:
+  // each is materialised, applied, and released before the next begins.
+  for (int repeat = 0; repeat < 10; ++repeat) {
+    b = *a_matrix * (u + v) + *b_matrix * (u - v);
+  }
```
icui: Is this mix usage of pointer and value intensional for testing?

Issue #2039, opened 2026-08-17
Reference (undeformed) GLL coordinate set for model sampling


lsawade:
*Milestone: 3D Globe I - basic global implementation*
**Depends on:** #1995, #2038
**Blocks:** #2043

## Context

This is the single most easily-missed correctness requirement in the globe integration, and
it is worth stating precisely.

The globe mesher samples the physical model on the **spherical, Moho-stretched** mesh —
*before* topography and ellipticity move the points. From
`src/meshfem3D/compute_element_properties.f90`, the per-element order is:

```
Moho stretching → interpolate to GLL → get_model (pass 2) → topography
  → 410/660 internal topography → ellipticity → recompute Jacobian
```

If SPECFEM++ called the oracle at the **final deformed** positions, near-surface points
would sample the wrong depth (topography is kilometres) and ellipticity would shift
latitude (about 20 km of flattening). The result is a mesh whose geometry is right and whose
material is subtly wrong everywhere near the surface — hard to spot, and it corrupts every
seismogram.

The writer therefore emits two anchor sets (record 11, present when
`HAS_REFERENCE_GEOMETRY`):

```
11  xref(nnode), yref(nnode), zref(nnode)   [SI metres, spherical + Moho-stretched]
```

PR #2011 captures these after Moho and radial-layer stretching but before surface
topography, ellipticity, and model-specific internal topography, and exposes
`specfempp_has_reference_geometry()` to decide when the record is required.

SPECFEM++ must interpolate the reference anchors to its GLL points with **the same shape
functions** used for the final geometry, reproducing the mesher's
`compute_element_GLL_locations`, and call the oracle there.

## Task description

- [ ] Add a container holding reference GLL coordinates — either a
      `reference_points` mixin on `assembly::mesh<dim3>` alongside `points`, or a standalone
      sibling struct. Prefer whichever keeps `assembly::mesh`'s mixin list honest; a
      standalone container is fine since nothing but property build reads it.
- [ ] Populate it by contracting the **same** shape functions used in issue #2038 against the
      reference anchors from record 11.
- [ ] Allocate it only when the header's `HAS_REFERENCE_GEOMETRY` is set. When unset,
      reference coordinates are by definition identical to final ones, so consumers should
      fall back to `points` rather than pay for a duplicate copy.
- [ ] Expose the fallback behind one accessor (e.g. `model_sampling_coordinates()`) so
[...]

PR #2097, opened 2026-09-21
Issue 2039 - Reference (undeformed) GLL coordinates for model sampling

lsawade:
The globe mesher samples the physical model on the spherical, Moho-stretched mesh —
before topography and ellipticity move the points. Sampling at the final deformed
positions would silently corrupt near-surface material. This PR adds the reference
coordinate set to `assembly::mesh<dim3>`:

- New `reference_points<dim3>` mixin holding reference GLL coordinates, built only
  when the database header sets `HAS_REFERENCE_GEOMETRY`. It shares the global
  numbering views with the final points and contracts the **same shape functions**
  against the reference anchors from the database.
- `model_sampling_coordinates()` is the single access point: it returns the
  reference set, or the final points when no reference geometry exists (identical
  by definition). Marked `@warning`: model sampling only — never for the Jacobian,
  mass matrix, or anything geometric.
- `read_globe_properties` now samples the oracle at these coordinates instead of
  interpolating anchors inline.

Decisions:
- The reference container is **kept alive** for the assembly's lifetime rather than
  released after property build.
- Interpolation accumulates in `type_real` instead of the previous `double`; anchors
  are already stored as `type_real`, so the difference is metre-scale at Earth radius
  in float builds — negligible for model sampling.

Tests: reference vs final coordinates differ by the fixture's ellipticity (0.5–40 km
radial); the accessor falls back to final points without reference geometry; identical
anchors reproduce the final coordinates to round-off; the reader round-trips the
reference record. The 1e-6 spot check against mesher-dumped GLL values is deferred to
a follow-up issue

## Issue Number

Closes #2039

## Checklist


tests/unit-tests/assembly/dim3/mesh/reference_points.cpp
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  {
+    SCOPED_TRACE("ellipticity separates reference and final coordinates");
+    const auto max_radial_difference = test_impl::max_radial_difference(
+        mesh_with_reference.model_sampling_coordinates(), final_points);
+    // The fixture has no topography, so the difference is the ellipticity
+    // deformation: kilometre scale at the surface, bounded by the flattening.
+    EXPECT_GT(max_radial_difference, 500.0);
+    EXPECT_LT(max_radial_difference, 40000.0);
```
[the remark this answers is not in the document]
lsawade: not applicable

core/specfem/assembly/assembly/dim3/assembly.cpp
```diff
@@ -59,6 +59,21 @@ make_element_types(
            mesh.globe.element_context };
 }
 
+specfem::mesh::control_nodes<
+    specfem::element::dimension_tag::dim3>::CoordinatesViewType
+reference_anchor_coordinates(const specfem::mesh::cartesian3d_mesh &) {
```
Rohit-Kakodkar, line 64: Should we not return coordinates view as reference coordinates when they are the same? In cartesian reference is the same as the deformed right?

core/specfem/assembly/assembly/dim3/assembly.cpp
```diff
@@ -59,6 +59,21 @@ make_element_types(
...
+  return {};
+}
+
+specfem::mesh::control_nodes<
+    specfem::element::dimension_tag::dim3>::CoordinatesViewType
+reference_anchor_coordinates(const specfem::mesh::globe3d_mesh &mesh) {
+  if (!mesh.globe.has_reference_geometry) {
+    return {};
```
Rohit-Kakodkar, line 72: Here too.

core/specfem/assembly/mesh/dim3/impl/reference_points.hpp
```diff
@@ -0,0 +1,73 @@
...
+ * stores their interpolation to the GLL points, computed with the same shape
+ * functions and element ordering as the final geometry.
+ *
+ * When no reference anchors are given, the container stays empty and
+ * consumers fall back to the final points (which are then identical by
+ * definition).
+ *
+ * @see specfem::assembly::mesh_impl::points
+ */
+template <> struct reference_points<specfem::element::dimension_tag::dim3> {
```
Rohit-Kakodkar, line 27: Do we need another struct for reference_points? This seems like a thin wrapper around ``specfem::assembly::points``. Apart from a boolean value this struct doesnt add any extra information.

Rohit-Kakodkar (changes requested):
``reference_gll_points`` should just be of type ``specfem::assembly::points``. Based on whether the mesh provides a separate reference points section we can decide to create a new view or point this to original points value. Will that not suffice?

core/specfem/assembly/mesh/dim3/impl/reference_points.hpp
lsawade: the problem is the the inheritance pattern of assembly. The alternative is to give assembly an optional member that is `specfem::assembly::points reference_points`. I went with the inheritance it was a choice.

core/specfem/assembly/assembly/dim3/assembly.cpp
lsawade: I do like that, then the reference_points should just be a member that is always set.

lsawade:
@Rohit-Kakodkar will refactor!

lsawade:
Updated after this morning's discussion.

icui (approved):

